PR #1, opened 2015-08-12
Ansible take 2

manics:
Currently just roles for setting up a jenkins-slave node including a storage partition.

Config files (host_vars, group_vars, host inventories) are in a separate private location.


manics:
This was successfully used to setup the latest jenkins node:

    ansible-playbook -i xxx-hosts -u XXX -b --ask-become-pass -v provision.yml

Then on Jenkins ssh to the new node, accept key, quit without entering password. And add the new node to Jenkins.

sbesson:
Tested locally the logic to combine the public tasks with private configuration files:

```
~/venv/bin/ansible-playbook -i /path/to/hosts -u XXX -b --ask-become-pass -v /path/to/public/provision.yml 
````

Logic seems to work as expected. Assuming @kennygillen signs off, let's get this merged and build on top of that.

PR #3, opened 2015-08-21
Centos7 omero build

manics:
This should setup almost all dependencies according to https://trello.com/c/SQPUIolT/257-python-2-7-on-centos-ci-nodes and @rleigh-dundee for building OMERO, including the Jenkins slave. This may need a development Ansible instead of the latest release (depending on which roles you test). README updated. I haven't actually tested whether OMERO runs/builds on this.

I've also moved all defaults variables into role defaults so `{group,host}_vars/all` is now empty (See http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable for variable precedence)

## Testing
See the README, in particular the last example for testing the setup of a Centos-7 build node. PostgreSQL on Centos-7 uses systemd by default which isn't properly supported in Docker, so there are some conditionals to get it working if the `runningindocker` variable is set. This means it should be possible to test almost all of this PR in Docker, with a final test in a VM to check PostgreSQL works.

Example inventory for testing in docker:
```
[ci-jenkins-linux]
omero-jenkins-slave-temp.docker runningindocker=True
```

## Things to note
- All yum tasks use `state: latest` which means everytime the playbook is run the packages will be updated. The alternative would be to change this to `state: present` and have the upgrade tasks in a separate role/playbook.
- Ice is currently hardcoded to 3.5 and the RPMs are installed directly in the interests of saving time (the multi-ice archive hasn't yet been setup).
- QT5 is not setup
- TexLive is not setup
- I'll try and add some contributing guidelines in a later PR for people to robustly argue over.
- There's a `ci-deployment.yml` playbook, with the roles split up as they are you should be able to create a production runtime-only playbook with only essential dependencies.
-  Multi-Ice, QT5 and TexLive can be setup in a future PR, or if it's vital now it could be manually installed and replaced later, especially since with Ansible we can easily wipe and rebuild the node :)

/cc @kennethgillen @joshmoore @sbesson


kennethgillen:
:+1: Good to merge. Nice work!

kennethgillen:
```bash
[khgillen@infra-testpr ~]$ sudo su - postgres
[sudo] password for khgillen:
-bash-4.2$ psql
psql (9.4.4)
Type "help" for help.

postgres=# \l
                                  List of databases
   Name    |  Owner   | Encoding |   Collate   |    Ctype    |   Access privileges
-----------+----------+----------+-------------+-------------+-----------------------
 postgres  | postgres | UTF8     | en_US.UTF-8 | en_US.UTF-8 |
 template0 | postgres | UTF8     | en_US.UTF-8 | en_US.UTF-8 | =c/postgres          +
           |          |          |             |             | postgres=CTc/postgres
 template1 | postgres | UTF8     | en_US.UTF-8 | en_US.UTF-8 | =c/postgres          +
           |          |          |             |             | postgres=CTc/postgres
```

Issue #6, opened 2015-08-27
Travis-CI syntax check


manics:
Something along the lines of:
```
for playbook in ansible/*.yml; do
    ansible-playbook --syntax-check $playbook
done
```

PR #7, opened 2015-08-27
Add Travis syntax check

sbesson:
Closes https://github.com/openmicroscopy/infrastructure/issues/6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+sudo: false
+
+install:
+  - pip install ansible
+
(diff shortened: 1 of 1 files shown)


sbesson:
Note this script currently assumes all playbooks will be stored directly under `ansible/`. If we are planning a different layout, it should be possible to run the syntax check on all `*.yml` files excluding `main.yml` /cc @kennethgillen @manics @joshmoore

joshmoore:
Works for me for the moment.

PR #11, opened 2015-09-24
prefixed the vars, added resizefs, since first post PXE task is alway…

kennethgillen:
…s resize the PXE created LVs and FS.

Feel free to test with infra_testpr. There's a snapshot ready for reverting-to (name of snap is self documenting).

PR #12, opened 2015-09-24
Remove Jenkins storage, update lv vars

manics:
Replace `jenkins-storage` with the generic `lvm_partition` role. Add `lvm_` prefix to vars in playbooks following #11

--depends on #11


sbesson:
Generally looks sensible to me. Leaving @kennethgillen to have a look at the code changs.

sbesson:
Tested on infratest-pr with a development anasible. Worked as expected.

PR #14, opened 2015-09-29
Install git on IDR nodes

manics:
Add `versioncontrol-utils` role for servers which need to manage code but are not full development/build servers. Also changed `yum state` from `latest` to `present` to avoid upgrading some packages such as docker without warning. If there's no objections I'll do the same for all other roles in a future PR.

Tested with `idr-deployment.yml` (i.e. I ran the playbook and applied the changes).

### DIFF
--- a/ansible/idr-deployment.yml
+++ b/ansible/idr-deployment.yml
@@ -7,3 +7,4 @@
   - role: docker-dns-server
   - role: docker-dns-client
   - role: samba-client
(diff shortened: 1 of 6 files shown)


joshmoore:
:+1: for git on narwhal

joshmoore:
That's a crazy number of files for something like this .... :smile:

PR #18, opened 2015-09-29
Add choice of either Ice 3.5 or 3.6

manics:
This has already been run against cowfish.

### DIFF
--- a/ansible/ci-deployment.yml
+++ b/ansible/ci-deployment.yml
@@ -12,3 +12,4 @@
   vars:
     - javajdkversions:
       - 1.8.0
+    - ice_version: 3.6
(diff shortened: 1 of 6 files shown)


joshmoore:
Did @sbesson have any worries about this? Thought I heard a snippet of something.

manics:
Apparently it wasn't needed, but we'll want it at some point.

sbesson:
As @manics said, it is not needed immediately (but was tested since Ice 3.6 was installed and had to be reverted to install Ice 3.5). Since the plan is to go through a clean reinstall of cowfish at some point, could we push an extra commit to change the default back to 3.5 for `ci-deployment.yml`?

manics:
Oops, I made the 3.5 change but got distracted

sbesson:
Tested by removing the Ice packages and the repo on cowfish, running ansible-playbook with the current state of the PR and re-deploying a server (https://ci.openmicroscopy.org/view/DEV/job/OMERO-DEV-Python27-merge-deploy/17/console). Will merge this once Travis is green.

PR #22, opened 2015-10-09
QA Web playbook

sbesson:
This PR defines a new playbook for deploying nodes used for the Web applications (registry, QA). Depending on their usage (staging/production), they can be used with/without `ci-provision.yml` to hook them to Jenkins. /cc @kennethgillen

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -58,9 +58,10 @@ As another alternative you can probably install ansible into your virtualenv (un
 Examples
 --------
 
-In the following examples replace example-hosts with the private host inventory file
+In the following examples replace example-hosts with the private host
+inventory file
 
-Dry-run `provision.yml` for all hosts listed in `provision.yml`:
+Dry-run `ci-provision.yml` for all hosts listed in `ci-provision.yml`:
 - `-u` Login as this user
 - `-b` use `sudo`
 - `--ask-become-pass` prompt for sudo password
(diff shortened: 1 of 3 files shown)


ansible/README.md
```diff
@@ -58,9 +58,10 @@ As another alternative you can probably install ansible into your virtualenv (un
 Examples
 --------
 
-In the following examples replace example-hosts with the private host inventory file
+In the following examples replace example-hosts with the private host
+inventory file
```
manics: Line wars! I've been working on the basis that inserting line-breaks is a pain.

sbesson: If that's the only issue with this PR, I am happy to revert it

sbesson: https://github.com/sbesson/infrastructure/commit/e3b79ab03d8cf715227aa541da2b612d7f0b4e45

manics: Cheers. Didn't want to get into a situation where everyone's doing different things. Rest of the PR looks OK

manics:
Ooops...the nginx role wasn't tested properly when it was merged :) It relies on `yum install epel-release` having been previously run, so it should work fine on our LSC VMs which might have epel enabled, but not on the docker container I used for testing. Apart from that it works fine.

What do you think- install `epel-release`? Or deal with separately along with the other clean-ups/improvements?

sbesson:
I think the main target of this new playbook are LSC VMS right @kennethgillen? If so, I would defer the installation/enabling of `epel-release` to a cleanup PR.

manics:
ok, want me to merge then?

sbesson:
I assume this will be tested on the new QA Web VMs that will be spinned up so I defer to Kenny no whether we should merge now or later.

ansible/qa-web-deployment.yml
```diff
@@ -0,0 +1,10 @@
+---
+# Playbook for provisioning QA Web nodes
+
+- hosts: qa-web-linux
+  roles:
+    - role: postgresql
+    - role: nginx
+  vars:
+    - javajdkversions:
```
manics: You don't need these vars unless you're installing java

sbesson:
Merging. Provisioning issues to be opened.

PR #21, opened 2015-10-03
Orca changes/fixes

manics:
Changes/additions made whilst provisioning orca

PR #23, opened 2015-10-13
Install check-mk

manics:
Currently only narwhal, orca will have to wait for #21


joshmoore:
No immediate objections, but perhaps better for @kennethgillen to also sign-off?

sbesson:
Got thumbs up from @kennethgillen, merging.

PR #24, opened 2015-10-21
Nginx proxy role

manics:
This is for configuring `web-proxy` (I've bagsied it for the IDR)

### DIFF
--- a/ansible/roles/nginx-proxy/templates/nginx-confd-dynamic-proxy.j2
+++ b/ansible/roles/nginx-proxy/templates/nginx-confd-dynamic-proxy.j2
@@ -1,0 +1,58 @@
+server {
+    listen       80 default_server;
+{% if nginx_proxy_ssl %}
+    listen       443 ssl;
+{% endif %}
+    server_name  $hostname;
+
+{% if nginx_proxy_ssl %}
+{% if nginx_proxy_ssl and nginx_proxy_ssl_certificate is defined %}
+    ssl_certificate {{ nginx_proxy_ssl_certificate }};
+{% endif %}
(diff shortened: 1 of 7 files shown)


ansible/roles/nginx-proxy/templates/nginx-confd-dynamic-proxy.j2
```diff
@@ -0,0 +1,58 @@
...
+
+    # To force dynamic DNS lookups need to use a variable:
+    # https://www.jethrocarr.com/2013/11/02/nginx-reverse-proxies-and-dns-resolution/
+    # http://forum.nginx.org/read.php?2,215830,215832#msg-215832
+
+{% for item in nginx_dynamic_proxies %}
+    location {{ item.location }} {
+        set $backend_{{ item.name }} "{{ item.server }}";
+        proxy_pass http://$backend_{{ item.name }};
+        proxy_redirect http://$backend_{{ item.name }} http://$server_name;
```
aleksandra-tarkowska: should we update nginx templates? did you notice any mismatch in attempt to connect with HTTP to HTTPS ports, or vice versa?

manics: This is an nginx config for proxying a proxy, so it'll be different from the OMERO web templates.
If the frontend proxy is using `https` but the backend is using `http` you sometimes get a mismatch, but I think that could be handled by https://trello.com/c/7eFWt9Bj/78-django-secure-proxy-ssl-header

joshmoore:
Merging as the current state of web-proxy.

PR #28, opened 2015-11-06
Jenkinsslave improvements

### LINKED ISSUE #26: jenkinsslave: .gitconfig needs to be created
When run for the first time, the task fails with

```
fatal: [web-qa-staging.openmicroscopy.org]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_args": {"dest": "/home/hudson/.gitconfig", "group": "hudson", "owner": "hudson", "src": "gitconfig.j2"}, "module_name": "template"}, "msg": "file (/home/hudson/.gitconfig) is absent, cannot continue", "path": "/home/hudson/.gitconfig", "state": "absent"}
```

### LINKED ISSUE #27: jenkinsslave: install git
This should likely belong to the tasks when setting up a Jenkins slave.

sbesson:
Fixes #26. Fixes #27

### DIFF
--- a/ansible/roles/jenkinsslave/tasks/setup_virtualenv.yml
+++ b/ansible/roles/jenkinsslave/tasks/setup_virtualenv.yml
@@ -1,0 +1,15 @@
+---
+
+- name: virtualenv | install python tools
+  yum:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - python-pip
+    - python-virtualenv
+    
+- name: virtualenv | create
+  pip: virtualenv=/home/{{ jenkinsuser }}/virtualenv
+
+- name: Ensure scc is installed
+  pip: virtualenv=/home/{{ jenkinsuser }}/virtualenv name=scc
--- a/ansible/roles/jenkinsslave/tasks/setup_virtualenv.yml
+++ b/ansible/roles/jenkinsslave/tasks/setup_virtualenv.yml
@@ -8,8 +8,5 @@
     - python-pip
     - python-virtualenv
     
-- name: virtualenv | create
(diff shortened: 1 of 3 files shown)


ansible/roles/jenkinsslave/tasks/setup_virtualenv.yml
```diff
@@ -0,0 +1,12 @@
...
+- name: virtualenv | install python tools
+  yum:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - python-pip
+    - python-virtualenv
+    
+- name: virtualenv | scc
+  pip: virtualenv=/home/{{ jenkinsuser }}/virtualenv name=scc
```
manics: Can you use the long-form?
```
pip:
  virtualenv: /home/{{ jenkinsuser }}/virtualenv
  ...
```

sbesson:
@manics pushed

manics:
Looks fine.

In general I think we should try and move most of the package setup into `ci-deployment` so it can be tested in docker, and leave `ci-provision` for hardware configuration which can only be tested in a VM. Downside is some tasks e.g. setting up the jenkins partition would be separate from setting up jenkins. Merge and leave this for later?

sbesson:
The main reason why I didn't include it into `ci-deployment` is that the latter contains omero specific tasks. Otherwise, I am fine with migrating the `virtualenv` part into `ci-deployment`.
Time to split the playbooks into something like :

```
ci-provision - CI hardware
ci-deployment - CI environment (SSH key, virtualenv)
omero-deployment - OMERO build deployment
docs-deployment -Docs deployment
```

Can do so in a follow-up PR?

PR #30, opened 2015-11-10
installing python-devel

aleksandra-tarkowska:
installing python-devel

### DIFF
--- a/ansible/roles/python-devel/tasks/main.yml
+++ b/ansible/roles/python-devel/tasks/main.yml
@@ -1,12 +1,7 @@
 ---
 # tasks file for roles/python-devel
 
-- name: check default python version
-  shell: "python --version 2>&1 | awk '{print $2}'"
-  register: python_installed_version
-
 - name: install python-devel
   sudo: True
   yum: name={{ item }}  state=present
   with_items: [ "python-devel", "gcc" ]
-  when: python_installed_version.stdout | version_compare('3.0', '<')
--- a/ansible/roles/python-devel/tasks/main.yml
+++ b/ansible/roles/python-devel/tasks/main.yml
@@ -1,0 +1,12 @@
+---
+# tasks file for roles/python-devel
+
+- name: check default python version
+  shell: "python --version 2>&1 | awk '{print $2}'"
+  register: python_installed_version
+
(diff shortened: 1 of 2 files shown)


ansible/roles/python-devel/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+# tasks file for roles/python-devel
+
+- name: check default python version
+  shell: "python --version 2>&1 | awk '{print $2}'"
+  register: python_installed_version
```
manics: We should be in full control of these machines, so the python verison should be known.

aleksandra-tarkowska:
@manics like that?

sbesson:
Works for me. @kennethgillen @manics: okay to give this a try?

ansible/roles/python-devel/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+# tasks file for roles/python-devel
+
+- name: install python-devel
+  sudo: True
+  yum: name={{ item }}  state=present
+  with_items: [ "python-devel", "gcc" ]
```
manics: Only remaining thing is a style issue- can you use the long form (makes it easier to review diffs in future), see e.g. https://github.com/openmicroscopy/infrastructure/blob/master/ansible/roles/omero-basedeps/tasks/main.yml#L9


and `sudo`, we should be consistent (add it to all required tasks?)

aleksandra-tarkowska:
Done

ansible/roles/python-devel/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+# tasks file for roles/python-devel
+
+- name: install python-devel
+  yum:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - "python-devel"
```
manics: Very pedantic...don't need quotes, but I'll let @sbesson have the final say

sbesson: This can be done in a separate PR. Still trying to come up with a workflow to test these PRs using a Docker with systemctl. Can be discussed at 11am.

PR #34, opened 2015-11-19
Add role for creating nginx self-signed certificates

manics:
This is useful for testing https internally.


joshmoore:
My attempts at testing this in docker (mgmt 80) didn't lead to a completely successful web-proxy with SSL, but the test did run through to completion in a ssh-systemd container. I'd propose merging and we incrementally add on to this since it's an internal debugging role rather than a production one.

kennethgillen:
:+1:

PR #35, opened 2015-11-20
Docker pg

joshmoore:
Role and playbook for running a production postgres instance in docker. It's intended that any host in the "docker-pg" group will have a postgres:9.4 image spin up on it and will listen on the 5432 docker host. No provision is made for running multiple postgres instances since it's assumed that a number of services will share the same postgres cluster.


manics:
> No provision is made for running multiple postgres instances

What does this mean? Couldn't you still run multiple pg containers with different `docker_pg_port`?

joshmoore:
> What does this mean? Couldn't you still run multiple pg containers with different docker_pg_port?

Mentioned to @manics by voice that this is intended for creating shared postgres instances. Happy to have the variables in place to let the role be added twice, but doesn't seem a priority.

sbesson:
discussed today, More work coming.

ansible/group_vars/all
```diff
@@ -1 +1,2 @@
 # Default / placeholder for variables
+docker_use_tls: verify
```
manics: My general thoughs on `groupvars`/`hostvars` under infrastructure were not to use them, since they'll need to be duplicated in our private configs. Could you move this into a default role variable instead, or even make it mandatory if it's not likely to break anything?

joshmoore:
Closing for the moment.

PR #39, opened 2015-11-30
Install LSB core package on all nodes

rleigh-dundee:
Addition of a single package.


manics:
This isn't part of the centos7 docker image, so assuming it works I think it should go in regardless of whether it's included in the standard VM images.

manics:
Or another alternative:
- `basedeps` has base dependencies
- `nonvmbase` has packages which are in the standard VM image that aren't in Docker.

sbesson:
Discussed at the sysadmin meeting. Merging and deploying on cowfish in a tick

PR #42, opened 2016-01-25
Ansible tasks: state: present, become: yes

manics:
- Use `state: latest` for all package installations apart from explicit upgrades, this will avoid uncessary checks and should prevent unexpected upgrades.
- Use `become: yes` for most tasks. It makes more sense to mark the tasks explicitly as requiring `sudo` instead of specifying it at the command line- this means you don't need the `-b` flag (though you may still need to pass `--ask-become-pass` to pass the sudo password).
- 2131636 uses `become: {{ jenkinsuser }}` for the creation of the virtualenv only
- Currently untested, thought I'd open it in case someone has time and to hopefully avoid merge conflicts

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,22 +65,21 @@ In the following examples replace example-hosts with the private host inventory 
 
 Dry-run `ci-provision.yml` for all hosts listed in `ci-provision.yml`:
 - `-u` Login as this user
-- `-b` use `sudo`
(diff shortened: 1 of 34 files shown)


manics:
FYI some of these role changes were tested whilst running https://github.com/openmicroscopy/infrastructure/pull/43

joshmoore:
Merging based on the testing of mgmt/99

PR #46, opened 2016-02-09
Network bonding, delete/disable unused NICs

manics:
Support configuration of bonded interfaces, see README for an example.
- `network_disable_ifaces` is to explicity create NIC config files to disable a NIC
- `network_delete_ifaces` is to delete unwanted `ifcfg-.*` files, for instance kickstarting a node may create an interface name which doesn't match the required naming.


joshmoore:
Tested on previous orca-2. Will be tested again after PXE'ing.

PR #47, opened 2016-02-10
Start refactoring CI playbooks

manics:
### Split CI-playbooks
- `ci-provision.yml` is now the run-everything playbook, and is made up of individual playbooks named after groups, which can be run independently. (Remember a host can be a member of multiple groups, so this allows you to manage one piece of "functionality" of a node)

- You can run `ci-provision.yml` with `--limit` to provision a new node, but if you are doing any hardware configuration it's a good idea to reboot in between. E.g. If you are reconfiguring the network run `ci-initial.yml` first, reboot, then run your other playbook, or just run the parent `ci-provision.yml` since Ansible is idempotent.

Note: I've started doing something similar with the IDR playbooks, but it made more sense to sort out ci-docker first. Also once everyone's happy with a workflow I'll add some docs (basically the above paragraphs).

### Tidied up some vars:

- There's no need to force `docker_basefs` `docker_lvfilesystem` to be defined when they could be set as defaults in the role.
- Added `root_filesystem` to idr playbooks for consistency.
- Made `network/network_ifaces` var optional, this means you can run `ci-initial.yml` for any node and if you haven't configured any sepcial networking the playbook will still run.

### DIFF
--- a/ansible/ci-docker-dev-hosts.yml
+++ b/ansible/ci-docker-dev-hosts.yml
@@ -14,6 +14,7 @@
     lvm_lvsize: "{{ scratch_size }}"
     lvm_lvfilesystem: "{{ scratch_filesystem }}"
   - role: basedeps
+#  - role: system-monitor-agent
   - role: docker
   - role: docker-dns-server
   - role: docker-dns-client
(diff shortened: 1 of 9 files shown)


joshmoore:
Local merge worked. Merging.

PR #51, opened 2016-02-16
Docker default filesystems

manics:
Small cleanup- make some mandatory vars optional.

Note `ansible/idr-provision.yml` group `idr-docker` is out of date (uses the old docker role) so probably best to ignore it.


joshmoore:
Local merge worked. Merging.

PR #53, opened 2016-03-03
Add playbook for simple upgrades of distro packages

manics:
Runs `yum update` on all hosts in the inventory (`all` is an ansible keyword meaning all hosts).

### DIFF
--- a/ansible/all-upgrade.yml
+++ b/ansible/all-upgrade.yml
@@ -1,0 +1,6 @@
+# Generic playbook for updating hosts which don't require any special tasks
+# For instance, this should not update excluded packages
+
+- hosts: all
+  roles:
+  - role: upgrade-distpackages


joshmoore:
```
TASK [upgrade-distpackages : system packages | upgrade] ************************
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks so that the environment value uses the full variable syntax ('{{upgrade_distpackages}}'). This
feature will be removed in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

If that's not a worry, I'd say merge as an independent playbook.

manics:
Looks like https://github.com/manics/infrastructure/blob/47bff976019729a9f9cd9732353ad14ede5049f1/ansible/roles/upgrade-distpackages/tasks/main.yml#L9 needs to be `{{ upgrade_distpackages }}`. Issue added: https://github.com/openmicroscopy/infrastructure/issues/55

joshmoore:
:+1:

PR #54, opened 2016-03-03
Add roles for connecting a VM to active directory, and for creating samba shares

manics:
`active-directory-join`: Take a clean CentOS 7 VM, run this with a host-vars file including AD username/password, and you should be able to ssh in with ldap.

`active-directory-samba-share`: See the readme for an example config for setting up samba shares. This is not dependent on `active-directory-join`, but it does assume AD has already been configured.

Docker might work too (untested).

### DIFF
--- a/ansible/idr-samba-hosts.yml
+++ b/ansible/idr-samba-hosts.yml
@@ -1,0 +1,6 @@
+---
+# Playbook for maintaining IDR Samba nodes
+
+- hosts: idr-samba-hosts
+  roles:
+  - role: active-directory-samba-share
--- a/ansible/roles/active-directory-join/README.md
+++ b/ansible/roles/active-directory-join/README.md
@@ -1,0 +1,27 @@
+Active Directory Join
+=====================
+
+Join a Linux server to an existing Active Directory domain.
+
+
+Role Variables
+--------------
+
+`active_directory_realm`: The main AD domain
+`active_directory_workgroup`: The samba workgroup
+`active_directory_server`: A list of the AD servers
+`active_directory_realmd_tags`: A list of AD realm tags
+`active_directory_kerberos_realms`: A dict of the form `{REALM: [domain, ...]}` where keys are the realms and values are a list of domains associated with that realm
+
+`active_directory_join_ou`: The domain/unit the server should be joined to
+`active_directory_join_user`: The `username%password` of the AD administrative account used to join the node
+`active_directory_join_access`: A list of user and/or group names that should be allowed to access this node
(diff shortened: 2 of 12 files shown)


joshmoore:
Trying to test on docker with aeb70a46d8ae7447cb117180594ae69196b48125 results in:

```
testsamba_1 | fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "no service or tool found for: smb"}
```

If that's not going to work, we may want to look into vagrant tests to reduce the testing burden. (There's also a lingering question of whether we should move to independent role repositories a la independent docker repositories, but that's a larger discussion for later).

manics:
Is systemd working?

joshmoore:
Switching to `omero-ssh-systemd` as the base image still shows the "no service" error.

manics:
Do any of the roles work? E.g.

`docker run ... name ... -it --rm -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v /run openmicroscopy/omero-ssh-systemd`

active-directory.hosts:
```
[samba-share-hosts]
test-ansible.docker.... ansible_user=omero
```

playbook.yml:
```
- hosts: samba-share-hosts
  roles:
  - role: active-directory-samba-share
    active_directory_realm: AD.EXAMPLE.ORG
    active_directory_workgroup: workgroup
    active_directory_shares:
      share-ro:
        path: /srv/samba/ro
        comment: Read-only share
        users: [root]

      share-rw:
        path: /srv/samba/rw
        comment: Read-write share
        readonly: False
        users: [root]
```

From a different host:
```
[root@11709d985723 /]# smbclient -L //simon-ansible.docker.openmicroscopy.org
Enter GUEST's password:
Anonymous login successful
Domain=[WORKGROUP] OS=[Windows 6.1] Server=[Samba 4.2.3]

	Sharename       Type      Comment
	---------       ----      -------
	share-ro        Disk      Read-only share
	share-rw        Disk      Read-write share
	IPC$            IPC       IPC Service (Samba Server Version 4.2.3)
Anonymous login successful
Domain=[WORKGROUP] OS=[Windows 6.1] Server=[Samba 4.2.3]

	Server               Comment
	---------            -------

	Workgroup            Master
	---------            -------
```
Logins won't work because AD isn't configured in this example, but you should still see a list of shares.

joshmoore:
Added simple samba test (from https://github.com/openmicroscopy/infrastructure/pull/54#issuecomment-194978579) in https://github.com/joshmoore/infrastructure/commit/dfe35dbfd7ca85d2fc72e233b4b3eae882dceba0 to gh-58.

joshmoore:
Note: I never got the docker-compose container working. Happy to come back to this as systemd evolves in docker. Otherwise, we can stay with vagrant.

PR #57, opened 2016-03-10
Use committed docker.service instead of copying/patching default.

manics:
Previously the docker role copied the distributed `/usr/lib/systemd/system/docker.service` into `/etc/systemd/system/` (which overrides the distributed files), then patched it. The intention was to keep abreast of any upstream changes in docker.service.

In practice this just adds complexity, so this commit bundles the full `docker.service` file to be used. Be aware that if docker is upgraded and `/usr/lib/systemd/system/docker.service` is changed you will need to merge in changes into `roles/docker/files/docker.service`.

One consequence of this was that the systemd startup timeout was not modified after an upgrade (`TimeoutStartSec=0` in `docker.service`), which caused the docker daemon startup to timeout on a heavily used system.


joshmoore:
Testing based https://github.com/openmicroscopy/infrastructure/pull/58 -- :thumbsup:

PR #59, opened 2016-03-18
Add some initial documentation for ansible

manics:
Some docs for discussion in next week's sysadmin/infrastructure/omero-deploy/docker meeting

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 Tools, scripts and other systems infrastructure used to support the work of the OME http://www.openmicroscopy.org/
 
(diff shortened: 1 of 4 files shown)


joshmoore:
Re: earlier conversation -- worth adding points re: "recommended-versions only", etc?

manics:
Pushed

joshmoore:
+1 thanks. Certainly a good starting point. How much feedback to you want pre-merging?

manics:
If there's no pressure to merge at the moment you might as well give as much feedback as you want, and if/when it grows too much we can discuss.

PR #62, opened 2016-04-04
Openstack installation helpers and docs

manics:
Contains some roles and changes to help with setting up OpenStack hypervisor nodes. Also some docs on setting up and using OpenStack.

### DIFF
--- a/openstack/idr-openstack-setup-tenant.sh
+++ b/openstack/idr-openstack-setup-tenant.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+# Setup default networks and security groups for an OpenStack tenant
+
+set -eu
+
+TENANT=${1:-}
+SUBNET=${2:-}
+DNS1=${3:-}
+DNS2=${4:-}
+if [ -z "$TENANT" -o -z "$SUBNET" -o -z "$DNS1" ]; then
+    echo "ERROR: Usage: `basename $0` NEW-TENANT SUBNET DNS1 [DNS2]"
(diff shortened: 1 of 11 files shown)


openstack/idr-openstack-setup-tenant.sh
```diff
@@ -0,0 +1,40 @@
...
+# Create a router to connect the external and private networks
+
+neutron router-create ${TENANT}_router
+neutron router-gateway-set ${TENANT}_router external_network
+neutron router-interface-add ${TENANT}_router ${TENANT}_subnet
+
+# Create some security groups
+
+nova secgroup-create ssh "SSH from anywhere"
+nova secgroup-add-rule ssh tcp 22 22 0.0.0.0/0
```
joshmoore: This results in:

```
Set gateway for router idrprod_router
Added interface 9001a01d-8323-4a79-b3bb-0d68d30de12e to router idrprod_router.
+--------------------------------------+------+-------------------+
| Id                                   | Name | Description       |
+--------------------------------------+------+-------------------+
| a458ec32-06a8-4659-973c-c8d5676d481c | ssh  | SSH from anywhere |
+--------------------------------------+------+-------------------+
ERROR (NoUniqueMatch): Multiple security group matches found for name 'ssh', use an ID to be more specific.
```

joshmoore: My mistake. This came from using the wrong tenancy openrc file since the script itself doesn't create the tenancy.

joshmoore:
I've tested tenancy creation now (see `idrstaging`).  I'm going to merge this so we can work on restructuring in preparation for tagging idr-0.0.4 (or 0.1.0!)

PR #65, opened 2016-04-14
Add NFS server shares role

manics:
Old style NFS (no authentication)

There's a bug with the current NFS package which means you may error messages in the logs, and there may be delays when you first access a file https://bugzilla.redhat.com/show_bug.cgi?id=1285097

### DIFF
--- a/ansible/idr-nfs-hosts.yml
+++ b/ansible/idr-nfs-hosts.yml
@@ -1,0 +1,6 @@
+---
+# Playbook for maintaining IDR NFS nodes
+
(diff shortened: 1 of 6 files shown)


joshmoore:
I assume this is already deployed somewhere?

manics:
Runnign this on a non-up-to-date system it was necessary to manually run `systemctl start rpcbind` before `systemctl start nfs-server`. TODO: If this still occurs after a full package upgrade it should be included in this role.

manics:
NFS now starts automatically after a system upgrade.

PR #66, opened 2016-04-19
ice 3.6.3

jburel:
Bump ice version and download rpm
unfortunately the previous approach does not work.
tested via https://ci.openmicroscopy.org/view/Mgmt/job/ICE-centos6-archive/33/

### DIFF
--- a/ci/packages/ice-multi-archive/create_ice_multi_zip.sh
+++ b/ci/packages/ice-multi-archive/create_ice_multi_zip.sh
@@ -77,6 +77,12 @@ wget https://zeroc.com/download/rpm/el6/x86_64/libice-java-3.6.2-1.el6.x86_64.rp
 wget https://zeroc.com/download/rpm/el6/x86_64/libice3.6-c%2B%2B-3.6.2-1.el6.x86_64.rpm
 wget https://zeroc.com/download/rpm/el6/x86_64/libicestorm3.6-3.6.2-1.el6.x86_64.rpm
 wget https://zeroc.com/download/rpm/el6/x86_64/mcpp-devel-2.7.2-3ice.el6.x86_64.rpm
+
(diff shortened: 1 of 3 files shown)


manics:
Since docker let's us setup individual test environments could we get rid of the whole multi-ice package?

jburel:
Maybe after the release we could look into that, release is fast approaching and probably better to follow a workflow that worked for us before.

manics:
Ice 3.6 uses a different installation procedure for Python (pip instead of rpm), so I so think the old workflow would require modification anyway.

jburel:
I know, Let's discuss tomorrow. trying to minimise the changes.

jburel:
This is the approach I had to take due to a change on zeroc side. Will recheck post release if it is still necessary. ice 3.6.2 has been deployed on the nodes using the management job.

joshmoore:
Any reason not to merge, @jburel ?

jburel:
Before merging I might try to see if they have fixed the issue encountered at the time with zeroc website so we don't have to do ``wget`` for all the dependencies

jburel:
No luck, Something has changed on their site.
So we will have to use the change made in this PR

joshmoore:
Checking if this is tested regularly by a merge job.

joshmoore:
Job is pointing to J-m's repo. Suggest merging and updating the job.

sbesson:
To be reconsidered as part of the Ice 3.6.3 testing epic.

jburel:
bump to 3.6.3, still have to be download the rpm one at a time

jburel:
tested via https://ci.openmicroscopy.org/view/Mgmt/job/ICE-centos6-archive/40/

sbesson:
To be merged once the corresponding code PR is included into the OMERO code.

joshmoore:
No vetoes.

PR #70, opened 2016-04-29
openstack playbooks

joshmoore:
This removes the requirement on checking out a branch from @manics as well as the bootstrap script. This means the benefits described in the README are no longer possible, but with the introduction of the dynamic inventory I think this is balanced out. The general workflow is actually quite nice:

 * `ansible-playbook ... os-create.yml`
 * `ansible-playbook ... os-omero.yml`
 * ... any other playbooks ...
 * ... do your testing ...
 * `ansible-playbook ... os-delete.yml`

I'd even like to see this runnable by devspace in a "ci" tenancy.

Questions:
 * [x] At this point should `openstack/*.md` moves to `docs` and `openstack/examples/*.yml` move to `ansible`?
 * [ ] With https://github.com/openmicroscopy/infrastructure/pull/70/commits/41ecc027bf1ceadd033905dd5997b37d60c75d0a, can 00_openstack.sh be reduced/removed?
 * [ ] Remove `omero` usage from the `os-*.yml` files?


openstack/examples/README.md
```diff
@@ -11,71 +11,44 @@ Guide for the Impatient
...
+    ansible-playbook os-delete.yml
 
+Inventory
+---------
 
-Updating an OpenStack VM
-------------------------
-
-You can take advantage of an [Ansible dynamic inventory script for OpenStack (`openstack.py`)](http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-openstack-external-inventory-script) instead of having to manage an inventory file when using Ansible to push out changes.
+This direectory takes advantage of an
```
mtbc: directory

openstack/examples/README.md
```diff
@@ -11,71 +11,44 @@ Guide for the Impatient
...
-Then run:
-
-    ansible-playbook idr-os-playbook.yml
+You can override variables at the command line, for example (note double quoting is necessary if spaces are present):
 
-Alternatively you can override variables at the command line, for example (note double quoting is necessary if spaces are present):
-
-    ansible-playbook idr-os-playbook.yml -e "omero_vm_key_name='My Keypair'"
+    ansible-playbook os-create.yml -e omero_vm_name=FOO -e omero_vm_key_name=YOURKEY \
+        -e "omero_vm_flavour='My Keypair'"
```
manics: `omero_vm_flavour='My Keypair'`: change the param value

joshmoore: ...ok, but this is from you! :smile:

joshmoore: Nope. I'm just wrong.

openstack/examples/README.md
```diff
@@ -11,71 +11,44 @@ Guide for the Impatient
...
-
-    ssh centos@IP_ADDRESS tail -f /var/log/cloud-init-output.log
-    ssh centos@IP_ADDRESS sudo journalctl -f
-
-The advantage of running this playbook from inside the VM instead of pushing it out is that no SSH authentication needs to be setup, and it is not necessary to wait for the VM to initialise its SSH server.
-
 To delete the VM and related security group:
 
-    ansible-playbook idr-os-playbook-delete.yml
+    ansible-playbook os-delete.yml
```
manics: Since you're editing this doc, worth mentioning that deleting the security group will fail (and can be ignored) if anything else is using that group?

openstack/examples/README.md
```diff
@@ -11,71 +11,44 @@ Guide for the Impatient
...
-------------------------
-
-You can take advantage of an [Ansible dynamic inventory script for OpenStack (`openstack.py`)](http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-openstack-external-inventory-script) instead of having to manage an inventory file when using Ansible to push out changes.
+This directory takes advantage of an
+[Ansible dynamic inventory script for OpenStack (`openstack.py`)](http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-openstack-external-inventory-script)
+instead of having to manage an inventory file when using Ansible to push out changes.
 For example:
 
-    ANSIBLE_ROLES_PATH=../../ansible/roles ansible-playbook -i openstack.py -l test-idr-omero idr-example-omero.yml -u centos -vv
+    ANSIBLE_ROLES_PATH=../../ansible/roles ansible-playbook -i inventory -l centos os-omero.yml -vv
```
manics: `-l centos`
Should this be the `omero` group? `-l omero`

openstack/examples/os-create.yml
```diff
@@ -5,15 +5,18 @@
...
   vars:
 
-    omero_vm_name: test-idr-omero
-    omero_vm_key_name: SSH keypair name
     omero_vm_image: CentOS 7.2 1602
-    omero_vm_flavour: m1.medium
-
-    setup_script: idr-example-omero-bootstrap.sh
+    omero_vm_flavour: m1.large
+    omero_vm_groups: "centos,omero"
```
manics: Thoughts on making it clearer these are groups, e.g. `centos-vms`, `omero-instances`, etc, or something else?

openstack/examples/os-omero.yml
```diff
@@ -31,6 +31,7 @@
     become: yes
 
   roles:
+  - role: versioncontrol-utils
   - role: upgrade-distpackages
```
manics: Very pedantic aesthetic point... I think `upgrade-distpackages` should be first:
1. Create an new VM
2. Upgrade existing packages
3. Install new stuff

manics:
:+1:, a few small comments inline.

FYI I was thinking of an all-in-one playbook for the IDR to spin up one or more VMs (e.g. separate postgres, I think this is needed to avoid future performance/debugging issues), and to run the required playbook roles- I think it's possible with the [add_host](http://docs.ansible.com/ansible/add_host_module.html). But that's for the a future PR

manics:
One more, fix the playbook name at the at the top of `os-delete.yml`: `# Delete VMs and security groups created by idr-os-playbook.yml`

openstack/examples/README.md
joshmoore: Yeah, I looked into: "if unused" but didn't find anything. Will push.

openstack/examples/README.md
joshmoore: Could probably use some discussion on that. The reason that it's `-l centos` is because that chooses `ansible_ssh_user` in `inventory/variables`.

openstack/examples/os-create.yml
joshmoore: Can do. But will wait on the above group discussion.

openstack/examples/README.md
manics:
```
diff --git a/openstack/examples/inventory/variables b/openstack/examples/inventory/variables
index fb175ba..36ab2c6 100644
--- a/openstack/examples/inventory/variables
+++ b/openstack/examples/inventory/variables
@@ -1,4 +1,4 @@
-[centos]
+[centos-vms]

-[centos:vars]
+[centos-vms:vars]
 ansible_ssh_user=centos
diff --git a/openstack/examples/os-create.yml b/openstack/examples/os-create.yml
index 62dd17b..0ad8c98 100644
--- a/openstack/examples/os-create.yml
+++ b/openstack/examples/os-create.yml
@@ -7,7 +7,7 @@

     omero_vm_image: CentOS 7.2 1602
     omero_vm_flavour: m1.large
-    omero_vm_groups: "centos,omero"
+    omero_vm_groups: "centos-vms,omero-vms"

   tasks:
```

`ANSIBLE_ROLES_PATH=../../ansible/roles ansible-playbook -i inventory -l omero-vms os-omero.yml -vv` seems to work

joshmoore:
Pushed a first round of fixes. Need to discuss groups first.

joshmoore:
Pushed better group names. Done unless you have suggestions.

joshmoore:
> FYI I was thinking of an all-in-one playbook for the IDR to spin up one or more VMs ... I think it's possible with the add_host. But that's for a future PR

Same here. I was assuming with an include an add_host wouldn't be necessary, but isn't so bad if it is.

joshmoore:
`add_host` example in place, using files migrated from https://gist.github.com/manics/d37fef6dcb7d3f946fb177aedf79b3c2

manics:
`~centos/idr-metadata` git checkout is owned by `root`, but I can't see why.

manics:
Style fixes in https://github.com/joshmoore/infrastructure/pull/1

manics:
@joshmoore Are you planning anything else?

joshmoore:
I don't have anything in flight. My top hope was to get the volumes auto-mounted but I failed to cleanly get the device name from os_server_volume (i.e. whether it already exists or not). We could go for a map of device names to volume names, but I haven't started on that.

manics:
Another strange problem: limiting to `-l localhost,FOO` (so that it won't affect existing instances) doesn't work unless `FOO` already exists:

```
$ ANSIBLE_ROLES_PATH=../../ansible/roles ansible-playbook -i inventory uod-os-idr-omero.yml -vv -e omero_vm_name=FOO -e omero_vm_key_name=XXX --list-hosts -l localhost,FOO

No config file found; using defaults
```

joshmoore:
@manics : pushed my latest fixes and the proposed `git mv` structure from my description question.

manics:
One nit-picky comment: You've got a mix of long (multiline) and short form YAML in some playbooks. If it's easier I can fix it in my pending commits.

manics:
(Sorry, accidentally clicked close instead of comment)

manics:
https://github.com/joshmoore/infrastructure/compare/os-playbooks...manics:pr70?expand=1

Main changes (you should be able to fast forward):
- There are now three separate VMs (Postgres, OMERO, gateway/caching proxy).
- Replaced `omero_vm_name` with `vm_prefix` which is used to prefix the names of all three VMs (`-omero`, `-gateway`, `-database`)
- I've reverted 80130836403178ca2b52c0eaa463f13371792d88 for now, but you can limit hosts using ansible's `-l` parameter, e.g. `-l vm_prefix-*`, `-l vm_prefix-omero`.
- Added a wrapper script `inventory/openstack-private.py`. `inventory/openstack.py` only returns instances with floating IPs by default, you have to pass `--private` to see the private IPs which is necessary if you're running ansible via an ssh gateway host.

Still more to come

manics:
Forgot to say so far I've mostly been testing with `uod-os-idr-omero.yml`. `os-create.yml` contains `add_hosts` tasks which means the openstack dynamic inventory isn't strictly needed if you're setting up the whole system from scratch, however it is required if you want to run `os-omero.yml` separately. Also be aware that `os-create.yml` automatically sets the ssh proxy variables to go via the gateway if `single_floating_ip: True` which is essentially what we have on Embassy. Change this to `False` to return to the previous behaviour of one floating IP and direct ssh connections for every instance. If you''re just running `os-omero.yml` you'll need to configure those ssh proxy variables somewhere else.

joshmoore:
Merging this as is since things are getting complicated. Leaving @manics to figure out the next phase.

manics:
Refactored my non-openstack changes into two separate PRs:
- https://github.com/openmicroscopy/infrastructure/pull/74
- https://github.com/openmicroscopy/infrastructure/pull/75

PR #72, opened 2016-05-03
Nginx proxy changes

manics:
- Support dynamic (look up IP everytime) and static backend servers.
- Also some cleanups to make this more general, e.g. the backend scheme is now required instead of only supporting http.
- Optional support for HTTP2 configuration (requires Nginx 1.10) added.
- This role uses a different name for the `conf.d/proxy.conf` file, and removes the old one if present `conf.d/dynamic-proxy.conf`

To be considered:
- [x] Use backend URI as key instead of front-end (to allow copying of cache between proxies)
- [ ] Cache static
- [ ] Limit number of front-end connections
- [ ] Add header to force a cache refresh
- [ ] `Expires` header to force clients to cache data http://nginx.org/en/docs/http/ngx_http_headers_module.html#expires


ansible/roles/nginx-proxy/tasks/main.yml
```diff
@@ -11,13 +11,13 @@
   become: yes
   yum:
     name: nginx
-    state: presentt
+    state: present
```
joshmoore: Odd that this didn't trigger a warning.

ansible/roles/nginx-proxy/templates/nginx-confd-proxy.j2
```diff
@@ -1,17 +1,30 @@
 server {
     listen       80 default_server;
```
aleksandra-tarkowska: maybe split them to two block rather then use if, see https://github.com/openmicroscopy/management_tools/blob/master/ci/config/cowfish/nginx/conf.d/OMERO-DEV-multiweb.conf#L13

aleksandra-tarkowska: sorry you were faster then me typing

joshmoore:
@kennethgillen / @manics : can you review the open items here (in the description & in comments) and open cards/issues as necessary? Merging via https://github.com/openmicroscopy/infrastructure/pull/78

PR #73, opened 2016-05-10
Add GPFS doc

manics:
Some notes on managing GPFS.


joshmoore:
Having these public works for me, but I'd vote for:

 * links from the top-level README (as in mgmt_tools)
 * consistent naming, either all markdown files in their respective directories (like `storage/`) or flattened (like the ansible files)

joshmoore:
Merging (partially for IDR-0.0.5)

PR #75, opened 2016-05-13
Allow external connections to postgresql server

manics:
At the moment only local connections are allowed by default. Added optional parameters `postgresql_server_listen` `postgresql_server_auth` to allow external connections.

This role now assumes full control of `pg_hba.conf`, and also inserts `listen_addresses =` into `postgresql.conf` even if it's the default `localhost`, so you'll see changes against existing uses of this role even though there should be no functional change.


rleigh-dundee:
All looks reasonable to me.  No actual use of postgresql_server_auth yet, or will this come later?

manics:
See this branch: https://github.com/manics/infrastructure/tree/openstack-playbooks-multivm
https://github.com/manics/infrastructure/blob/openstack-playbooks-multivm/ansible/os-omero.yml#L10

I cherry-picked these commits since they're logically independent of the other work in that branch.

manics:
Closing to avoid conflict with #78

joshmoore:
Merging via #78

PR #76, opened 2016-05-13
Bump GPFS kernel version to 3.10.0-327.18.2.el7.x86_64

manics:
I upgraded one server and this was the latest kernel.

### DIFF
--- a/ansible/roles/gpfs/defaults/main.yml
+++ b/ansible/roles/gpfs/defaults/main.yml
@@ -9,7 +9,7 @@ gpfs_install: True
 gpfs_configure: gpfs_install
 
 # Build/Install a GPFS module for this kernel version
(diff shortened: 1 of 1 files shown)


joshmoore:
How do we determine if this will cause problems elsewhere? Is it used only at install? Or would any execution of the role on the machine upgrade the kernel?

manics:
It's used on whatever VM is used to build the gpfs kernel module. I should probably set that var in a host_vars file instead. By default it should fail if the current kernel version doesn't match the one defined in the gpfs role https://github.com/openmicroscopy/infrastructure/blob/99ecfd190763793d2d4fac59e3659ad15497f13f/ansible/roles/gpfs/tasks/gpfs-install.yml#L4

manics:
Don't care about this anymore.

PR #79, opened 2016-06-09
Add image migration scripts

joshmoore:
In order to move images between OS clouds,
several steps are required. This is currently
being used for providing a clone of the EBI
servers at UoD.


manics:
These commits switch to using the openstack client (recommended over the individual clients where possible http://docs.openstack.org/cli-reference/common/cli_install_openstack_command_line_clients.html), and parse the json output to check for completion:

https://github.com/manics/infrastructure/commit/ac1cb8383b46ad81076f23800fa0e627f0290248
https://github.com/manics/infrastructure/commit/ee024fab99cd50b0c8cf37875b0a14bbdb7760e5

joshmoore:
Pushed.

joshmoore:
Merging as the current state of things.

PR #80, opened 2016-06-10
Web proxy setup

manics:
- Change ci-initial.yml to ansible/all-initial.yml, and apply to `all`, so it can be used for setting up any new server.
- Change `idr-services.yml` to expand root and log volumes

### DIFF
--- a/ansible/all-initial.yml
+++ b/ansible/all-initial.yml
@@ -3,7 +3,7 @@
 # Note you should reboot after changing the network configuration, and
 # manually verify that the configuration is correct
 
-- hosts: ci
+- hosts: all
   roles:
   - role: upgrade-distpackages
   - role: network
--- a/ansible/ci-provision.yml
+++ b/ansible/ci-provision.yml
@@ -3,7 +3,7 @@
 
 # Initial networking and basic system configuration
 # A reboot may be required after this playbook
-- include: ci-initial.yml
+- include: all-initial.yml
 
 # Jenkins CI servers
 - include: ci-jenkins-linux.yml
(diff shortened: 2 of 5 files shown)


joshmoore:
Barring

```
ERROR! the file_name '/home/travis/build/openmicroscopy/infrastructure/ansible/ci-initial.yml' does not exist, or is not readable
```

this looks good. (Tested a `-C` run locally with ansible 2.1 which handled LVM much better.)

manics:
The problem now is
```
$ ansible-playbook -i ansible/example-hosts --syntax-check ansible/*.yml
ERROR! vars file vars/os-create-{{ os_cloud_provider | default('uod') }}.yml was not found
```

joshmoore:
Is there a fix for that?

manics:
Not without refactoring the playbooks, it looks like you can't have a conditional (`when:`) in `vars_files`

joshmoore:
Then exclude this playbook from travis?

manics:
I've just found this https://github.com/ansible/ansible/issues/10010#issuecomment-70647392
Not quite the same behaviour (since it'll also default to `-uod` if `os_cloud_provider` is invalid instead of only when it's empty)

manics:
Pushed, see the commit message for caveats

joshmoore:
No possibility of a `noop.yml`? If not, let's go with this, and we can see how often it bites us.

manics:
I could do that, but it means `os_cloud_provider` becomes mandatory whereas previously it defaults to `uod`. Reckon it's worth it?

joshmoore:
Hmmm.....don't know. Let's hold off and see.

PR #81, opened 2016-06-13
Remove old docker roles

manics:
The one server that used to use the old `docker-{advanced,storage}` roles has been upgraded to the new `docker` role.

Ideally `idr-docker-old.yml` would be subsumed by `idr-docker-localstorage.yml`, but due to a different LVM configuration on the old servders this isn't possible.


joshmoore:
> Ideally idr-docker-old.yml would be subsumed by idr-docker-localstorage.yml, but due to a different LVM configuration on the old servders this isn't possible.

So once we re-image narwhal this can be deleted?

joshmoore:
Running a local `-C` version against idr-hosts passes (all green).

manics:
> So once we re-image narwhal this can be deleted?

Yes

joshmoore:
:+1:

PR #82, opened 2016-06-13
Use vdb for omero volume, support custom hostgroups for different environments

manics:
- All volumes now use the raw disk without partitions (`vdb`)
- group_vars have been moved to the `os-idr` group, so that they can be copied and overriden easily
- See updated README

### DIFF
--- a/ansible/README-os-idr.md
+++ b/ansible/README-os-idr.md
@@ -1,20 +1,26 @@
 Openstack IDR Playbooks
 =======================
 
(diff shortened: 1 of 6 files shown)


joshmoore:
Can you apply the same fix for
```
ERROR! vars file vars/os-create-{{ os_cloud_provider | default('uod') }}.yml was not found
```
here?

manics:
I'll rebase it on https://github.com/openmicroscopy/infrastructure/pull/80

manics:
@joshmoore No need to rebase, restarted travis instead since it's fixed on master.

joshmoore:
:+1:

PR #83, opened 2016-07-05
Re org and addition of prod omero playbook

kennethgillen:
Reorganised top-level doc, which had heavy OpenStack focus.

Left generic details, a little description, and also copied the OpenStack specific README to a new doc.

Added new playbook for a production standalone OMERO server, `prod-omero.yml`, as the existing OMERO server playbook was the OpenStack specific os-omero.yml. 

Was a little more specific with the message/name in ` ansible/roles/omero-server/tasks/omero-systemd.yml` as it's trying to start up the servers too, not just enable them. I had some issues with running this with an already-running OMERO, so the specific 'enable and start' is a little more descriptive about what it's trying to do.


kennethgillen:
As for testing, the `prod-omero.yml` was used to bring up the server `training-server` so you likely don't need to re-verify it works if time short.

I'm mostly looking for vetoes on the new structure of the top level doc - which made this repo look like it was only for OpenStack, which might put off collaboration and reuse.

ansible/README.md
```diff
@@ -1,60 +1,13 @@
...
 Most of these scripts should also work on other platforms, providing the VM is brought up by some other method.
-The guest must be running CentOS 7.
 
+- Playbooks which start with `os` are OpenStack specific - the `os` stands for `openstack`.
+- Playbooks which start with `idr` are for the Image Data Repository.
+- Playbooks which start with `ci` are for OME continuous integration and build.
 
-Guide for the Impatient
------------------------
+HERE LINK TO: [[README-os.md]]
```
joshmoore: This isn't resolving under https://github.com/kennethgillen/infrastructure/tree/bce516f7d509371e87cdedfcced7ce01e25e88bc/ansible

ansible/README.md
```diff
@@ -1,60 +1,13 @@
...
-
-This directory takes advantage of an
-[Ansible dynamic inventory script for OpenStack (`openstack.py`)](http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-openstack-external-inventory-script)
-instead of having to manage an inventory file when using Ansible to push out changes.
-For example:
-
-    ansible-playbook -i inventory -l os-image-centos os-omero.yml -vv
-
-Variables for the groups defined in os-create.yml as omero_vm_groups can be added under inventory/variables.
+HERE LINK TO:
```
joshmoore: Drop?

ansible/README.md
```diff
@@ -1,60 +1,13 @@
...
+===========
 
-This contains example files for provisioning an OpenStack VM from scratch with OMERO using Ansible.
+This contains a variety of Ansible playbooks and roles, including example files for provisioning an OpenStack VM from scratch with OMERO using Ansible.
 Most of these scripts should also work on other platforms, providing the VM is brought up by some other method.
-The guest must be running CentOS 7.
 
+- Playbooks which start with `os` are OpenStack specific - the `os` stands for `openstack`.
+- Playbooks which start with `idr` are for the Image Data Repository.
+- Playbooks which start with `ci` are for OME continuous integration and build.
```
joshmoore: Add a line about what `prod-*` means?

joshmoore:
:+1: on the docs change barring the few comments. Primary question which may be for a later clean up is how far are we away from `os-omero.yml` being able to re-use `prod-omero.yml`? cc: @manics Otherwise, *nice*

Note: additionally tested with:

```
(v)~/opt/infrastructure/ansible $ git diff
diff --git a/ansible/Vagrantfile b/ansible/Vagrantfile
index 1974913..e06a490 100644
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -31,6 +31,14 @@ Vagrant.configure(2) do |config|
     end
   end

+  config.vm.define "prod-omero" do |prod|
+    prod.vm.box = "centos/7"
+    prod.vm.provision "ansible" do |ansible|
+      ansible.verbose = "v"
+      ansible.playbook = "prod-omero.yml"
+    end
+  end
+
```

(though I have had some issues running it)

manics:
I reckon keep them separate, the common functionality should be in the role

manics:
Otherwise you get into the problem of combinatorial testing- if you modify settings in the playbook you now need to check those settings don't affect any other playbooks which use it (e.g. overwriting/hiding variables)

joshmoore:
I think you have that testing problem anyway because if you change a role, you need to find & update each the locations.

manics:
I think it's more complicated with playbooks because you have to take into account group/host vars (which may or may not also be set in one or more playbooks)

joshmoore:
But assuming we reviewed the variables couldn't the upstreamers have fewer and the downstreamers have more overrides? Yes it's work, but it seems that that's the kind of refactoring that will keep all of this reusable.

ansible/README.md
kennethgillen: Yeah, sorry: was supposed to actually link to the files I'd effectively replaced by this general README. Updated now. Was in the "have to get a PR with the playbook in" to be able to comment on a PR with the output of the plays.. circular holding pattern..

joshmoore:
:+1:

PR #84, opened 2016-07-07
Role for running Dell updates

manics:
We've got enough identical servers that's it worth automating server updates. See example playbook in README.

@kennethgillen

### DIFF
--- a/ansible/roles/dell-update/README.md
+++ b/ansible/roles/dell-update/README.md
@@ -1,0 +1,62 @@
+Dell Update
+===========
+
+Run Dell Linux driver updates such as those for the BIOS and iDRAC.
+
+Assumes the update package takes the following options (`--help`):
+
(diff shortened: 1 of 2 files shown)


manics:
This isn't ideal since if the update triggers a reboot Ansible reports an error, amongst other things. However this is only intended to be run when you actively want to install an update since you probably don't want to be checking for drac/bios upgrades every day.

kennethgillen:
Since tested on the Orcas, good to merge - one we decide if it should be here, or in management_tools. 👍  and ➕ 1 for automating the process.

joshmoore:
Merging. More general discussion around the example playbooks. Propose: move example playbooks to separate files in infrastructure (if big) or into `test.yml` (if small). When roles are extracted to separate repos, that example should be extracted as well.

PR #87, opened 2016-07-14
Refactor IDR round 1

manics:
This is the first of several PRs involved in refactoring these playbooks and variables. Mostly involves moving and renaming.

- Rearranges the IDR baremetal playbooks into an `idrsystems` group (to distinguish it from the IDR itself)
- See README.md for explanation of the two toplevel `idrsystems-*yml` playbooks


manics:
One thing I'm uncertain about is whether `idrsystems-playbook/idr-initial.yml` should actually be a top-level standalone playbook, since in most cases you should reboot the server after initialising the network card and updating system packages.

joshmoore:
Merging as a first proposal for new directory names. I'll update gh-88 to match this structure and add onto the proposal. @manics will open the mgmt docs PR once PXE'ing is done.

PR #89, opened 2016-07-21
New docker nodes, add ganesha NFS, update GPFS

manics:
Multiple changes related to setting up some new servers, and re-imaging the broken GPFS/NFS server.

## `roles/docker`
- Option added to use a different repo, e.g. `testing` to install docker release candidates
- Fixed variable quoting warning

## `roles/gpfs`
- Added a simpler alternative to the current build and install steps. If you have access to a GPFS yum repository, and you don't mind installing a compiler tool-chain on your server, you can install the base RPMs from the repo, compile the kernel module on the server, and install it

## `roles/nfs-ganesha-share`
- The `nfs-share` role uses the built-in kernel NFS server, however this has problems when serving GPFS: https://access.redhat.com/solutions/1553943. nfs-ganesha is a user-mode server which may work better.
- Note that this is the default `VFS` module, not the specialised `GPFS` nfs-ganesha module which is probably more efficient and reliable. This will be investigated in future.

## `roles/upgrade-distpackages`
- I tried to optionally have the role reboot the computer following a kernel upgrade. This works with VMs but may not work with physical servers where it takes longer than the initial delay (15s) to shutdown.

## `roles/nfs-mount`
- Mounts a list of NFS shares, this is easier than writing multiple `tasks:mount` in a playbook

### DIFF
--- a/ansible/idr-playbooks/idr-dundee-nfs.yml
+++ b/ansible/idr-playbooks/idr-dundee-nfs.yml
@@ -28,10 +28,10 @@
   - name: Mount NFS share
     become: yes
     mount:
-      fstype: nfs
+      fstype: nfs4
       name: "{{ idr_mountpoint }}"
       opts: rsize=8192,wsize=8192,timeo=14,intr,ro
-      src: idr-data.openmicroscopy.org:/uod/idr
+      src: idr-data.openmicroscopy.org:/idr
       state: mounted
     when: not use_samba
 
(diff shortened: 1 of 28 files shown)


manics:
@openmicroscopy/devops I used this branch to deploy the latest servers, but do you want me to split this into 4 PRs for review?

joshmoore:
Tested via PR 166

PR #88, opened 2016-07-14
Prod openstack (infra)

joshmoore:
Few cleanup commits as well as a few new playbooks for getting new production services on openstack:

 * `training-server` migrated from management_tools (thanks @kennethgillen!)
 * playbooks for production docker containers
 * playbooks for development web server

PR #94, opened 2016-08-11
Development website playbook

sbesson:
Ongoing work following https://github.com/openmicroscopy/infrastructure/pull/88 on a playbook for deploying a Jekyll-based website /cc @manics @kennethgillen


ansible/uod-web.yml
```diff
@@ -14,35 +14,62 @@
     live_site_aliases:
       - openmicroscopy.org
   roles:
-    - nginx
+    - role: versioncontrol-utils
+    - role: nginx
+    - role: jenkinsslave
   pre_tasks:
```
manics: Any reason for using `pre_tasks` as opposed to putting it all in `tasks`?

ansible/uod-web.yml
```diff
@@ -14,35 +14,62 @@
...
+      yum:
+        name: "{{ item }}"
+        state: present
+      with_items:
+       - rubygems
+       - ruby-devel
+       - make
+       - gcc
   tasks:
+    - name: Install Bundler
```
manics: Can you convert to long form YAML (and same for the rest of the playbook)?

manics:
@openmicroscopy/devops Does anyone have thoughts on putting things into subdirs vs putting them all into the ansible dir?

sbesson:
Re https://github.com/openmicroscopy/infrastructure/pull/94#issuecomment-239819851, would the proposal be:

```
ansibles/playbooks/subdir1
ansibles/playbooks/subdir2
...
```

or rather

```
ansibles/subdir1/<playbook>.yml
ansibles/subdir1/host_vars/
...
ansibles/subdir2/<playbook>.yml
ansibles/subdir2/inventory/
...
```

and would you match the set of `subdir` with the listing of https://github.com/openmicroscopy/management_tools/tree/master/ansible?

manics:
The general idea would be to try and push most of the common functionality into roles, and in future this is what we'd publicise to external people.

This means playbooks could be tailored for our specific servers, which they already are to a certain extent, especially the ones which create or modify LVMs, and putting them into subdirectories makes it clearer what playbooks should be run to manage particular servers. See e.g. the current `idrsytems-playbooks/` (there are current two top level playbooks, `idrsystems-deployment.yml` `idrsystems-provision.yml` which call the individual playbooks in `idrsytems-playbooks/*`, and I'm doing a similar refactoring with the idr openstack playbooks: https://github.com/openmicroscopy/infrastructure/compare/master...manics:refactor-idr.

Depending on how big the web work is you could either have a separate web directories, or make it part of a combined ci directory. Or if you prefer wait till more people are using it and there's a clearer idea of requirements.

sbesson:
Discussed at SA meeting, refactoring of the playbook into roles might happen as we use it in many servers. Merging for now as the current state is deployed in production.

PR #99, opened 2016-08-16
Rename idr-docker-localstorage to idr-docker

manics:
I need to split up the configuration of the `idr-docker-localstorage` group into two (the same variable needs to take different values on different nodes), and now that the old docker nodes have been re-imaged it's a good time to rename this group to just `idr-docker`


sbesson:
Discussed at sysadmin. Merging.

PR #100, opened 2016-08-17
install Pillow 2.9

aleksandra-tarkowska:
This fixes issues with Pillow

see output https://gist.github.com/aleksandra-tarkowska/11dd9cc0d8ec7b70f289fe33f7460f21
test training server deployed on http://10.0.51.137

slackscreensnapz001

### DIFF
--- a/ansible/roles/omero-web-runtime/tasks/main.yml
+++ b/ansible/roles/omero-web-runtime/tasks/main.yml
@@ -8,7 +8,6 @@
     state: present
   with_items:
     - python-pip
-    - python-pillow
 
 - name: omero web | install python redis package
   become: yes
@@ -25,6 +24,7 @@
     state: absent
   with_items:
     - python-gunicorn
+    - python-pillow
 
 - name: omero web | pip install packages
   become: yes
@@ -32,6 +32,7 @@
     name: "{{ item }}"
     state: present
   with_items:
+    - "Pillow==2.9"
     - "django>=1.8,<1.9"
     - "gunicorn>=19.3"
 
--- a/ansible/roles/basedeps/tasks/main.yml
+++ b/ansible/roles/basedeps/tasks/main.yml
@@ -20,3 +20,4 @@
     - unzip
     - wget
     - zip
+    - gcc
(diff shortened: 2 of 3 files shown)


kennethgillen:
Certainly looks good. Will test agains infra-testpr.

ansible/roles/omero-web-runtime/tasks/main.yml
```diff
@@ -25,13 +24,15 @@
     state: absent
   with_items:
     - python-gunicorn
+    - python-pillow
```
joshmoore: What's this line for?

joshmoore:
Are there any options for this that would prevent the `gcc` installation?

ansible/roles/omero-web-runtime/tasks/main.yml
aleksandra-tarkowska: to remove default one

joshmoore: Ah, missed  "state: absent". Thanks.

aleksandra-tarkowska:
the only option is to give a param to disable scripts. then python-pillow will be chosen. Note pip installed pillow is needed to run script which is pretty much server dependency. I was also planning to refactor and decouple web (as manic suggested omero-python role).

manics:
Have you tested this on a native openstack cento7 image?

joshmoore:
Another question: will any changes in the training playbook be needed re: https://trello.com/c/wEh5uadW/2134-re-do-pillow-on-frog-o-org?

aleksandra-tarkowska:
@joshmoore I am not sure if I understand the question. This PR is to fix training playbook `training-server.yml`. I understand that will affect others, but as it is possible to disable scripts optionally, then person running playbook will be able to decide. I don;t know any other option to install pillow 2.9 without gcc.

aleksandra-tarkowska:
Heads up. If we are not happy for bcc on prod then how you want to pop install zeroc-icepy ? https://github.com/ome/omero-install/blob/develop/linux/step01_centos7_ice_deps.sh#L23

joshmoore:
Think we're into gcc/g++ regardless. Might just be something to make clear upfront.

joshmoore:
This is being migrated to gh-105

PR #101, opened 2016-08-17
pip install the latest ice 3.6.x

manics:
Previously this would pip-install `ice 3.6.1`. It should now install the most recent in the `3.6` series.


joshmoore:
@aleksandra-tarkowska : any objections to merging this before we finish testing gh-105?

aleksandra-tarkowska:
that is ok, we are not installing that anyway

joshmoore:
:+1: Merging along with gh-102 & gh-103. Any cleanup can happen afterwards.

Issue #77, opened 2016-05-17
Non-idempotent tasks in omero-server role


manics:
https://github.com/openmicroscopy/infrastructure/pull/70 Introduced a non-idempotent pair-of tasks: https://github.com/openmicroscopy/infrastructure/blob/de1d03ae64511202401b1d370cc4d8f01c0d6c45/ansible/roles/omero-server/tasks/main.yml#L38

- `name: omero | copy bash skeleton files` copies `.bashrc` from `/etc/skel/`
- `name: omero | add to path` modifies the same `.bashrc`

PR #102, opened 2016-08-17
Omero server cleanups

manics:
- Fixes the lack of idempotency when modifying `~omero/.bashrc`: closes https://github.com/openmicroscopy/infrastructure/issues/77
- The `omero-server` role takes into account the value of `ice_version` from `roles/ice` (currently still `3.5`, should it be bumped to `3.6`?)
- Remove a harmless (though distracting) red ignored error message when checking for the existing version of a non-existent server
- `training-server.yml` updated to take account of the above

### DIFF
--- a/ansible/roles/omero-server/README.md
+++ b/ansible/roles/omero-server/README.md
@@ -26,6 +26,7 @@ This defaults to `latest`, but due to the broken registry a proper upgrade check
 You are advised to change this to an actual release version.
 - `postgresql_users_databases`: Used by the `postgres` role, if the PostgreSQL server is on the same node as the OMERO.server define this variable to create the local databases and users.
 If a remote database server is used a user and empty database must already exist.
+- `ice_version`: This variable originates from the `ice` role, leave unset for the default.
 
 
 Warning
--- a/ansible/roles/omero-server/README.md
+++ b/ansible/roles/omero-server/README.md
@@ -21,7 +21,7 @@ Role Variables
 
 All variables are optional, see `defaults/main.yml` for the full list
 
-- `release`: The OMERO release, e.g. `5.2.2`.
+- `omero_release`: The OMERO release, e.g. `5.2.2`.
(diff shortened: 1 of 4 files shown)


joshmoore:
Changes make sense. What running servers would be most affected by these changes so that we can run some playbook (possible `-C`) and see what the effects are?

manics:
It's shouldn't change anything on existing servers

manics:
Maybe frog since that's where `training-server.yml` was first run?

joshmoore:
> It's shouldn't change anything on existing servers

That's my point! :smile:

> Maybe frog since that's where training-server.yml was first run?

frog is a good idea if @kennethgillen allows it.

kennethgillen:
@pwalczysko wants to end up with a new golden OMERO db based on `frog`, so if you promise not to destroy the OMERO database or repository....

joshmoore:
Ran with `--check` (`-C`) and hit a failure:

```
TASK [training-server | add symlinks to sample images in fm users homedirs] ****
failed: [training-server.openmicroscopy.org] (item=fm1) => {"failed": true, "item": "fm1", "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /var/sample-images", "path": "/home/fm1/sample-images", "src": "/var/sample-images", "state": "absent"}
failed: [training-server.openmicroscopy.org] (item=fm2) => {"failed": true, "item": "fm2", "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /var/sample-images", "path": "/home/fm2/sample-images", "src": "/var/sample-images", "state": "absent"}

NO MORE HOSTS LEFT *************************************************************
        to retry, use: --limit @training-server.retry

training-server.openmicroscopy.org : ok=62   changed=8    unreachable=0    failed=1
```

This doesn't seem related to this PR, but it also isn't very comforting.

joshmoore:
Merging along with gh-101 and gh-103.

PR #103, opened 2016-08-18
Refactor idr production playbooks part 1

manics:
- Move `os-*yml` IDR related playbooks to `idr-playbooks/`
- Use a combination of trying multiple locations relative to the inventory file, and hardcoded paths, to avoid the need to symlink in multiple paths
- Remove `README-os.md` since it was mostly related to the IDR and moved relevant information into `README-os-idr.md`
- Attempt to fix the handling of command-line boolean vars in `os-idr-playbooks/os-create.yml` (it's not entirely clear whether `-e xxx=True/False` is treated as a bool or a string)
- Add some new vars and move some not-really-private vars into `idr-playbooks/idr-omero.yml` to allow the playbook to be run without the private config repo. This tests the basic infrastructure but won't result in the IDR since critical OMERO config isn't included.
- Add `idr-playbooks/idr-omero.yml` to `Vagrantfile`

Note this includes the commits from https://github.com/openmicroscopy/infrastructure/pull/102 so merge that first to keep things tidy.

See corresponding config PR for more details


joshmoore:
Reading through this, it generally makes sense. I'll test the new inventory mechanism as part of spinning up `spider`. Cheers.

PR #107, opened 2016-08-26
spiders: small modifications

joshmoore:
* decrease the size of the image to medium;
   small would possibly be better if ci-omero is removed
 * change the ordering of docker imports;
   the docker mount though is *still* not being mounted


joshmoore:
Merging as the state of the spiders. We'd likely like to see them as `m1.small` but that's ok for the moment.

PR #112, opened 2016-09-07
Update ansible.cfg for performance

joshmoore:
Most of the options included here are for working with
dozens of servers at once time. Also, they assume a
location for your local inventory (../../ansible/inventory)
which simplifies use for the OME team.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -9,9 +9,24 @@
 # Set `ANSIBLE_CONFIG` for local changes
 
 [defaults]
+# Try more servers at once, smartly.
+forks = 20
+gathering = smart
+
+# Assume all actions take place under infrastructure/ansible
 roles_path = roles/
+inventory = ../../ansible/inventory/
+
+# These tend to be annoying.
 retry_files_enabled = False
 
-# Allow becoming an unprivileged user
-# http://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user
-allow_world_readable_tmpfiles = True
+# Deals with the files/ directory used by IDR
+inventory_ignore_extensions = ~, .orig, .bak, .ini, .retry, .pyc, .pyo, .html, .omero, .gif, .png, .js, .md
+
+# Helps determine what's running slowly
+callback_whitelist = profile_tasks
+
+# Performance options
+[ssh_connection]
+ssh_args = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
+pipelining = True
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -28,5 +28,4 @@ callback_whitelist = profile_tasks
 
 # Performance options
 [ssh_connection]
-ssh_args = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
 pipelining = True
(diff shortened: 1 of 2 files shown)


sbesson:
Works for me. Only thought is that the `inventory` customization could likely be documented in the README if we are promoting these playbooks and expecting people to use them.

sbesson:
@kennethgillen: worth merging and including in the submodule bump for future seabass/tadpole work?

joshmoore:
> could likely be documented in the README if we are promoting these playbooks and expecting people to use them

I was hoping to hold off until we dropped the `ansible/` directory since "../../inventory" is probably a generally useful place to have your inventory (i.e. outside of the repo you just checked out). I'll go ahead and push a README commit though, and we can decide if it can be cleaned up.

joshmoore:
Note: when connection to openstack is blocked (hotel/VPN) `-i ../../ansible/inventory` may hang.

aleksandra-tarkowska:
This PR cause issues with running standard roles:

```
TASK [upgrade-distpackages : system packages | upgrade] ************************
task path: /Users/ola/OMERO/SYSADMIN/infrastructure/ansible/roles/upgrade-distpackages/tasks/main.yml:4
Monday 12 September 2016  14:46:05 +0100 (0:00:00.363)       0:00:00.403 ****** 
failed: [10.0.51.143] (item=[u'*']) => {"failed": true, "item": ["*"], "module_stderr": "sudo: sorry, you must have a tty to run sudo\n", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}

NO MORE HOSTS LEFT *************************************************************

10.0.51.143                : ok=1    changed=0    unreachable=0    failed=1   
```

see http://stackoverflow.com/questions/35597076/ansible-sudo-sorry-you-must-have-a-tty-to-run-sudo

joshmoore:
Definitely agreed though I haven't tracked down where *all* it requires fixing. My preference would be that we uncomment `require tty` in sudoers, but happy to see another proposal.

ansible/ansible.cfg
```diff
@@ -9,9 +9,24 @@
...
-allow_world_readable_tmpfiles = True
+# Deals with the files/ directory used by IDR
+inventory_ignore_extensions = ~, .orig, .bak, .ini, .retry, .pyc, .pyo, .html, .omero, .gif, .png, .js, .md
+
+# Helps determine what's running slowly
+callback_whitelist = profile_tasks
+
+# Performance options
+[ssh_connection]
+ssh_args = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
```
manics: `-o ForwardAgent=yes` means your ssh agent will be forwarded to all servers you use with Ansible, is this necessary? (Without this `-o ControlMaster=auto -o ControlPersist=60s` should be the default)

joshmoore: I'm drawing something of a blank, but I think the reason for adding this was for getting through the gateways. It might be that we want to add configuration at the host/group level for anything behind a gateway. In general, you and I should talk through the public/private split, since I'm sure I've left things in a surprising state for you.

joshmoore:
We should get at least parts of this in ASAP because without https://github.com/openmicroscopy/infrastructure/pull/112/files#diff-233067f858a1495efe2f23d1f1ed4548R24 the inventory is broken.

manics:
How about remove 
```
+[ssh_connection]
+ssh_args = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
```
then merge?

joshmoore:
Updated.

manics:
👍

PR #113, opened 2016-09-09
New seabass provision forced improvements

kennethgillen:
Improvements to the ci provision and deployment code, updated when adding a new OME ci slave. https://github.com/openmicroscopy/management_tools/pull/197

### DIFF
--- a/ansible/ci-deployment.yml
+++ b/ansible/ci-deployment.yml
@@ -6,6 +6,7 @@
     - role: jenkinsslave
     - role: sudoers
     - role: versioncontrol-utils
+    - role: system-monitor-agent
 
 # Deploy OMERO build and runtime prerequisites
 - hosts: ci-omero
--- a/ansible/ci-jenkins-linux.yml
+++ b/ansible/ci-jenkins-linux.yml
@@ -3,6 +3,13 @@
 
 - hosts: ci-jenkins-linux
   roles:
+  # Root LV Size 
+  - role: lvm-partition
+    lvm_lvname: root
+    lvm_lvmount: /
+    lvm_lvsize: "{{ provision_rootsize }}"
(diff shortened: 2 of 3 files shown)


joshmoore:
The LVM section _may_ need a `when not spacewalk` (`when baremetal`?) clause as well, since on openstack we haven't been using LVM (though maybe it won't hurt and we can start)

manics:
Something looks fishy here- you shouldn't need `enablerepo: epel`. If it's not spacewalk then epel should be enabled by default.

kennethgillen:
@manics: it seemed to be similar to your comment on ome-internal docs, which was that a freshly imaged system doesn't have any repo definition files, which it seems Ansible needs to read. epel is 'enabled' via spacewalk, but Ansible can't seem to understand that, and it fails, despite being able to install things from epel via yum.

```
After the first boot /etc/yum.repos.d/ is missing the standard repo files which can break Ansible
``` from `ome-internal/sysadmin/provisioning/index.html`

manics:
I thought LSC said you have to wait a bit until spacewalk updates the system?

kennethgillen:
@manics - even if that's the case, if we refactor the ansible code so we can bypass steps like "wait for system to refresh itself" then we might as well, no?

Note, `enablerepo: epel` was in there when I started, and if it is removed, and avoids this problem in the first place, then great. I was just writing a way around it, and assumed it was there for a reason, put there by someone else for a purpose.

kennethgillen:
> It was probably put there whilst I was trying to figure out how LSC/spacewalk worked (or not). Do you want to have a go at reverting this change and removing enablerepo: epel, or do you want someone else to?

No preference. I don't have time to work on that at the moment, so if anyone else wants to pick it up, then fair enough.

PR #105, opened 2016-08-20
improve ci deployment

aleksandra-tarkowska:
This PR improve ci slave deployment:
 - add epel repo
 - add hudson to sudoers
 - add omero postgres user and allow creating DB

# To test: first clone https://github.com/openmicroscopy/management_tools/pull/181
- define `management_tools/ansible/ci/ci-test-hosts`:

 ```
# Jenkins CI slave nodes
[ci-jenkins-linux]
ip1 (server)
ip2 (web)

[ci-omero]
ip1

[ci-omero-web]
ip2
```
- clone this
- cd infrastructure/ansible

- run playbook

 ```
ansible-playbook -i ../../management_tools/ansible/ci/ci-test-hosts  -vv -u centos ci-deployment.yml
```

Jenkins authorized key is set in https://github.com/openmicroscopy/management_tools/blob/master/ansible/ci/group_vars/ci-jenkins-linux#L8. make sure appropriate `SSH Username with private key` is chosen to Jenkins before connecting to the node.

see http://10.0.51.148/

Todo:
- no strategy for storing snoopy ssh keys

PR #119, opened 2016-09-22
add packages needed for selinux setup

hajaalin:
The packages I added are listed in
https://github.com/openmicroscopy/infrastructure/blob/master/ansible/roles/omero-server/tasks/ansible-prerequisites.yml
but that is included only after omero-web-runtime role is included as a dependency in 
https://github.com/openmicroscopy/infrastructure/blob/master/ansible/roles/omero-server/meta/main.yml.

I tested like this:
`ansible-playbook -i omero-deploy/inventory/dev infrastructure/ansible/training-server.yml`
with
`omero_selinux_setup: True`

With `omero_selinux_setup: False` nginx is not allowed to connect to OMERO-web at port 4080. This is on a Vagrant box centos/7.

### DIFF
--- a/ansible/roles/omero-server/defaults/main.yml
+++ b/ansible/roles/omero-server/defaults/main.yml
@@ -80,3 +80,10 @@ omero_systemd_setup: False
 # EXPERIMENTAL, may break your system
 # Quoted to prevent autoconversion to bool
 omero_systemd_restart: "no"
+
+# Setup trust store
+omero_trust_store_setup: False
+omero_trust_store: "{{ omero_basedir }}/trust_store.jks"
+omero_trust_store_passwd: omero
+#omero_trust_store_certificates:
+#  - { url: 'https://www.digicert.com/CACerts/DigiCertAssuredIDRootCA.crt', name: 'DigiCertAssuredIDRootCA.crt'}
--- a/ansible/roles/omero-server/handlers/main.yml
+++ b/ansible/roles/omero-server/handlers/main.yml
@@ -6,3 +6,8 @@
   service:
     name: nginx
     state: restarted
+
+- name: update config
+  become: yes
+  become_user: omero
+  shell: "{{ omero_serverdir }}/{{ omero_server_symlink }}/bin/omero load < {{ omero_basedir }}/config/omero-base.config"
--- a/ansible/roles/omero-server/handlers/main.yml
+++ b/ansible/roles/omero-server/handlers/main.yml
@@ -1,13 +1,27 @@
 ---
-# Handler for nginx
 
+# Handler for nginx
 - name: restart nginx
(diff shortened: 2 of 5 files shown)


manics:
Hi @hajaalin 

Thanks for your PR! We're part-way through decoupling the deployment of OMERO.web from OMERO.server, it looks like the duplication of the selinux tasks originates from #105. I'll discuss with the rest of the team how best to fix this.

manics:
@openmicroscopy/devops We should've picked this up with our [Vagrantfile](https://github.com/openmicroscopy/infrastructure/blob/master/ansible/Vagrantfile) :-(

manics:
@hajaalin I've cherry-picked your first commit into https://github.com/openmicroscopy/infrastructure/pull/124 since I required it for testing some of my other changes.

I'll leave 4e92796 unmerged for now, see the followup comments on #127

joshmoore:
@hajaalin : let us know when you want this to be reviewed again.

hajaalin:
@joshmoore : you can review it when you have time. It seems to work for me.

joshmoore:
Hi @hajaalin. Thanks again for this. We had a chance to go through your changes today. A couple of points:

 * Our fault for not having a style-guide in place, but we're trying to move toward the long-form (more yaml-like) rather than the short-form (a=b style on a single line). It makes reading the diff much easier.
 * How has your experience been with the restart handlers? We're struggling with them ourselves especially with regard to error conditions (i.e. will handlers run if a task fails?).

We're going to need some time from our side to have a solid proposal on the handlers front. If you'd like to split out just the trust store tasks, we'd like to go ahead and get that integrated. Alternatively, if you'd like to be more independent of us (at least initially) this likely could be it's own role. We're (very) slowly but surely going to break the roles out of this repo and move them to http://galaxy.ansible.com. Happy to hear your thoughts on that.

All the best, ~Josh.

manics:
@hajaalin I've just been through your Java trust store commit https://github.com/openmicroscopy/infrastructure/pull/119/commits/0bda40405c2990d21bac492b48161725c9a7327e. If I understand correctly it isn't OMERO specific, and could in fact be used by any Java application that requires a trust store. I think it would therefore be a good candidate for being made into an independent role- e.g. on Ansible Galaxy. Is this something you'd like to do?

hajaalin:
@manics Sorry for slow reply, holidays and such... Anyway, here's an attempt: https://galaxy.ansible.com/hajaalin/truststore/. I didn't try it with OMERO yet, but at least it passes some basic tests.

joshmoore:
Nice, @hajaalin. We're nearly ready to swap over to using galaxy ourselves (gh-239). Do let us know if you have any issues using all the roles together. Closing this in favor of your repo.

PR #120, opened 2016-09-23
Fix Vagrantfile so `vagrant up` brings up all nodes

manics:
The `idr-*` ansible config was previously global, including `ansible.limit = "idr"`. This PR makes it local to just the idr-* nodes now. `vagrant up` should now bring up all nodes.

Current HEAD is broken so tested against c5c56e34563b6dd3e3218b9176ba10ef21b235bd
You might be able to find a more recent working version.

### DIFF
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -51,22 +51,22 @@ Vagrant.configure(2) do |config|
       config.vm.provider :virtualbox do |vb|
         vb.customize ["modifyvm", :id, "--memory", "4096"]
       end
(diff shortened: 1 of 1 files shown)


manics:
First step to getting integration tests for this repo. vagrant has a [docker provider](https://www.vagrantup.com/docs/docker/configuration.html) which I'm looking into, which might allow us to run this on travis.

joshmoore:
Didn't try starting everything up completely, but file parsed and starting bringing up the IDR machines. Probably need to discuss what box could run this as an integration test.

PR #121, opened 2016-09-23
Return non-zero error code if no hosts match

manics:
This adds a callback so that Ansible will set a non-zero exit code if a playbook was run but no hosts match.  This is useful when running ansible-playbook automatically. For example, when running vagrant to test the Vagrantfile if there is a mismatch between the hosts expect by the playbook and the hosts in the Vagrantfile ansible would normally not care.

If you think this is inappropriate it could go into an alternative `ansible.cfg` which is used only by `vagrant` (e.g. set `ANSIBLE_CONFIG`)

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -26,6 +26,9 @@ inventory_ignore_extensions = ~, .orig, .bak, .ini, .retry, .pyc, .pyo, .html, .
 # Helps determine what's running slowly
 callback_whitelist = profile_tasks
 
+# Additional plugins
+callback_plugins = ./plugins/callback
+
 # Performance options
 [ssh_connection]
 pipelining = True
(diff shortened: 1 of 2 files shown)


joshmoore:
I get:

```
```

even without this PR. In fact, I don't I have yet been able to make it execute.

manics:
Run it without `--limit`, but with a playbook containing `- hosts: non-existent-group`

The idea is to catch errors in the tests, e.g. in the `IDR multinode omero` example https://github.com/openmicroscopy/infrastructure/pull/124 requires a prefix (`idr_environment`). If you were to change this but forget to change the prefix on the groups in the Vagrantfile ansible wouldn't match any hosts, but would still exit cleanly.

joshmoore:
```
$ /opt/v/bin/ansible-playbook none.yml

PLAY [DNE] *********************************************************************
skipping: no hosts matched

ERROR: No hosts matched
```

:thumbsup:

PR #123, opened 2016-09-26
Add optional cli utils

manics:
Add some generally useful utilities that aren't strictly necessary for a minimal production server.


joshmoore:
Is there a test in vagrant that would catch both the true and the false case here?

manics:
No. I'm wondering whether your original idea of splitting this into two roles makes for easier reading (simple declarations vs parsing a conditional):

    - role: basedep
    - role: cli-utils

vs

    - role: basedeps
      basedeps_user_utils: True/False

joshmoore:
Agreed that it would be good to come to a conclusion here:

 * new role requires update to playbooks; new conditional to variables
 * roles are easy to track but could multiply our public repos;  variables can be hard to track down since they can be changed on the command-line, environment, etc.

manics:
Now in a separate `cli-utils` role.

joshmoore:
Note: I'm getting

```
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

manics:
Try this on top or master (you'll need the fix from https://github.com/openmicroscopy/infrastructure/pull/120) since I've added commits instead of rebasing.

joshmoore:
Good tip. Cheers.

```
cliutils                   : ok=4    changed=3    unreachable=0    failed=0

Thursday 06 October 2016  10:43:26 +0200 (0:00:05.213)       0:04:04.990 ******
===============================================================================
basedeps : system packages | basic system utils ----------------------- 234.99s
cli-utils : system packages | cli user utils ---------------------------- 5.21s
basedeps : system packages | install epel repo -------------------------- 2.81s
setup ------------------------------------------------------------------- 1.92s
(v)~/opt/management_tools/infrastructure/ansible $ vagrant ssh cliutils
Last login: Thu Oct  6 03:23:58 2016 from 10.0.2.2
[vagrant@localhost ~]$ screen
[detached from 21316.pts-0.localhost]
[vagrant@localhost ~]$
```

PR #124, opened 2016-09-26
Get the IDR playbooks working smoothly with multiple deployments

manics:
Replaces `vm_prefix` with `idr_environment` in all openstack IDR playbooks, adds this as a group prefix in all IDR playbooks.

This means every server in an IDR deployment is in the following three groups
  1. one of `database-hosts` `omero-hosts` `gateway-hosts`
  2. `{{idr_environment}}-hosts`
  3. one of `{{idr_environment}}
-database-hosts` `{{idr_environment}}
-omero-hosts` `{{idr_environment}}
-gateway-hosts`

In practice this means it is possible to configure variables for
  1. all servers by purpose (database, omero, gateway)
  2. all servers in a given deployment (idr_environment)
  3. a server by purpose in a given deployment (intersection of the above)

At present (3) should only contain one server (which can be addressed as e.g. `{{idr_environment}}
-database-hosts[0]`, if we implement load-balancing there may be more than 1.

This also reverts some changes to `ansible.cfg` from https://github.com/openmicroscopy/infrastructure/pull/112 in order to get this repo working, these changes should be reverted in a `development` or `breaking` branch until everyone's happy.

`idr_environment` defaults to `idr` so that the IDR playbooks can be run without any private configuration. In most cases you should set `idr_environment` to something else and add matching group vars files. For example see `# IDR multinode omero` in Vagrantfile

Fixes the following from the config issue:

 * [x] tty required for sudo (causes `fatal: [demo3-omero]: FAILED! => {"failed": true, "msg": "Failed to get information on remote file (/etc/security/limits.d/95-omero.conf): MODULE FAILURE"}` (idr-omero.yml))
 * [x] use of `[0]` for which host to use is dangerous since it assumes one type per cloud.

Includes:
- https://github.com/openmicroscopy/infrastructure/pull/119 first commit only https://github.com/openmicroscopy/infrastructure/pull/119/commits/d25f4d15bf34504d362356a1fce9d94af5bf5591
- https://github.com/openmicroscopy/infrastructure/pull/120
- https://github.com/openmicroscopy/infrastructure/pull/123
- https://github.com/openmicroscopy/infrastructure/pull/125

For tidiness this PR should be rebased after merging the above three.

PR #128, opened 2016-09-27
Move most IDR config to this repo

manics:
Moves most of the IDR configuration into this public repo.
Adds placeholders for variables which should remain secret, where possible a public default is given:
```
$ git grep idr_secret
idr-playbooks/files/IDR-OMERO-52-omero.j2:config set omero.web.public.password {{ idr_secret_omero_web_public_password | default("") }}
idr-playbooks/group_vars/omero-hosts.yml:omero_system_uid: "{{ idr_secret_omero_system_uid | default(546) }}"
idr-playbooks/group_vars/omero-hosts.yml:omero_logmonitor_slack_token: "{{ idr_secret_omero_logmonitor_slack_token | default(None) }}"
```
Most of these playbooks should therefore run without any private configuration.
Note: work on changing `omero_system_uid` in the IDR is ongoing.

## Testing

Ansible group_vars can be in two locations, the playbook directory and the inventory directory, with the former taking precedence. This means when you run a playbook in `idr-playbooks` `idr-playbooks/group_vars` is automatically included, and if you also specify an inventory (`ansible-playbook -i INVENTORY/PATH`) `INVENTORY/PATH/group_vars` will be included.

I've only tested this on top of https://github.com/openmicroscopy/infrastructure/pull/124


manics:
What would be ideal is if we could automatically create a `public` user if it doesn't already exist, but that probably requires a custom omero ansible module.

joshmoore:
Merging for the upcoming testing via demo3 deployment.

PR #20, opened 2015-10-02
Add sphinx

ghost:
- Add docs role including sphinx
- Add genshi
- Add python role for use by any role using python pip/venvs

PR #129, opened 2016-09-27
Improve doc ci node

aleksandra-tarkowska:
add missing package(s):

- add `Pygments`, doc prerequisites
- add `ansible`, omeroweb-install prerequisites

see https://ci.openmicroscopy.org/job/TEST-DEV-merge-docs and https://ci.openmicroscopy.org/job/TEST-DEV-merge-docs-autogen

keep in mind https://github.com/openmicroscopy/infrastructure/pull/20

### DIFF
--- a/ansible/ci-deployment.yml
+++ b/ansible/ci-deployment.yml
@@ -7,7 +7,6 @@
     - role: sudoers
     - role: versioncontrol-utils
     - role: system-monitor-agent
-    - role: ansible
 
 # Deploy OMERO build and runtime prerequisites
 - hosts: ci-omero
@@ -37,4 +36,5 @@
 # Deploy Documentation prerequisites
 - hosts: ci-docs
   roles:
+    - role: ansible
     - role: texlive
--- a/ansible/ci-deployment.yml
+++ b/ansible/ci-deployment.yml
@@ -7,6 +7,7 @@
     - role: sudoers
     - role: versioncontrol-utils
     - role: system-monitor-agent
+    - role: ansible
 
 # Deploy OMERO build and runtime prerequisites
 - hosts: ci-omero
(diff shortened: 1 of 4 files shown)


sbesson:
As briefly mentioned yesterday, I think a series of next steps would be:
- remove `epydoc` (fully replaced by Sphinx to generate the Python API doc)
- make the `texlive` role optional and/or remove the role in the context of https://trello.com/c/SnUesylG/505-ditching-docs-pdfs
- refactor the Sphinx dependency handling out of `jenkinsslave` into a separate role. A constraint of the current docs build logic is the Sphinx + Ant dependency. In the context of building docs via Jenkins, this means we would either need Ant installed in the role (as proposed in the role https://github.com/openmicroscopy/infrastructure/pull/20), or Sphinx installed globally instead of using a virtualenv or some code work to remove this dependency.

manics:
> this means we would either need Ant installed in the role (as proposed in the role #20), or Sphinx installed globally instead of using a virtualenv or some code work to remove this dependency.

If we're moving to more compartmentalised build servers e.g. a docker container or VM that only builds docs everything could be installed at the global level if that makes things easier

aleksandra-tarkowska:
My intention is only to move away doc jobs from cowfish as server was overloaded. If we would like to do more complex rearrangement of the roles, maybe we could handle that separately. As the pdf removal decision hasn't been made. I could wait for @rleigh-codelibre return to help with fixing `texlive`

joshmoore:
spider-5 is now created:

```
[hudson@spider-5 ~]$ ./virtualenv/bin/pip freeze | grep Pyg
Pygments==1.4
[hudson@spider-5 ~]$ which ansible
/bin/ansible
```

I'm merging this and will open a doc PR against MT.

joshmoore:
NB: I assume it's possible to test moving PRs to this jenkins node.

PR #64, opened 2016-04-06
docker: add docker-compose installation to the role

joshmoore:
This currently installs globally via python-pip from EPEL.

PR #131, opened 2016-09-29
Add docker-tools role to install docker-compose globally

manics:
I need this for the IDR so created this role based on https://github.com/openmicroscopy/infrastructure/blob/master/ansible/roles/devspace/tasks/main.yml

If everyone's happy I'll update `roles/devspace` to depend on this.

/cc @aleksandra-tarkowska

### DIFF
--- a/ansible/roles/docker-tools/tasks/main.yml
+++ b/ansible/roles/docker-tools/tasks/main.yml
@@ -1,6 +1,12 @@
 ---
 # Install additional docker tools globally
 
+- name: system packages | install epel repo
+  become: yes
+  yum:
+    name: epel-release
+    state: present
+
 # Install as many python dependencies using RPMs if possible to reduce risk
 # of future conflicts
 - name: docker | Install docker python tools
--- a/ansible/roles/docker-tools/tasks/main.yml
+++ b/ansible/roles/docker-tools/tasks/main.yml
@@ -1,0 +1,31 @@
+---
+# Install additional docker tools globally
+
+# Install as many python dependencies using RPMs if possible to reduce risk
+# of future conflicts
+- name: docker | Install docker python tools
+  become: yes
+  yum:
+    name: "{{ item }}"
+    state: present
+  with_items:
+  - python-devel
+  - python-pip
+  - PyYAML
+  - python-six
+  - python-requests
+
+# TODO: Ansible fails with latest version of docker-py
+# Error: docker-py version is 1.10.2. Minimum version required is 1.7.0.
+# See https://github.com/ansible/ansible/issues/17495
+#
+# The latest docker-compose requires docker-py==1.10.3, so cap version
+# https://github.com/docker/compose/blob/1.8.1/requirements.txt#L4
+- name: docker swarm | install python packages
+  become: yes
+  pip:
+    name: "{{ item }}"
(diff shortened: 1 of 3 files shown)


ansible/roles/docker-tools/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+#
+# The latest docker-compose requires docker-py==1.10.3, so cap version
+# https://github.com/docker/compose/blob/1.8.1/requirements.txt#L4
+- name: docker swarm | install python packages
+  become: yes
+  pip:
+    name: "{{ item }}"
+    state: present
+  with_items:
+  - docker-py==1.9.0
```
aleksandra-tarkowska: I think docker-py==1.9 must go after compose, thats why I moved that to next task becuse it will be easier to remove once they fix it

aleksandra-tarkowska: and pip must be upgraded as well, see https://github.com/aleksandra-tarkowska/infrastructure/blob/3b7c0862a0388a867c1b58a36170e291b96bdddf/ansible/roles/devspace/tasks/main.yml#L41

joshmoore:
Interestingly, I had done this earlier today:

```
diff --git a/ansible/roles/docker/tasks/main.yml b/ansible/roles/docker/tasks/main.yml
index 7186aec..c0fd945 100644
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -11,8 +11,23 @@
 - name: upstream docker | install docker
   become: yes
   yum:
-    pkg: docker-engine
+    pkg: "{{ item }}"
     state: present
+  with_items:
+    - docker-engine
+    - python-devel
+    - python-pip
+
+- name: install python packages
+  become: yes
+  pip:
+      name: "{{ item }}"
+      #extra_args: "--upgrade"
+      state: latest
+  with_items:
+    - pip
+    - pyyaml
+    - docker-compose
```

manics:
@joshmoore From past experience with multiple dependencies automatically installed using pip (which you might not realise) I though it's better to keep it separate, especially if you've got more than just docker installed.

manics:
https://github.com/openmicroscopy/infrastructure/pull/64

joshmoore:
No objections, but I don't understand what should be kept separate from what.

manics:
Whenever you install anything outside yum there's a risk of installing something that'll conflict, especially pip which installs python dependencies which can conflict with distro python-module rpms

aleksandra-tarkowska:
@joshmoore docker-py has to be installed after docker-compose and capped to 1.9.0 https://github.com/ansible/ansible/issues/17495

manics:
@aleksandra-tarkowska It works for me, see https://github.com/openmicroscopy/infrastructure/pull/131/commits/8f8f5ba04170e53ee0a2eb9f4de0639d6f421157

@joshmoore Ideally this would go into a virtualenv, but then you need to figure out how to get ansible to pick it up.

aleksandra-tarkowska:
@manics sorry I missed the fact that you capped compose on 1.8.0.

joshmoore:
@manics is no longer using this strategy, but it's useful for devspace.

PR #134, opened 2016-10-03
Exe4j: license

sbesson:
Noticed by @jburel during OMERO 5.2.6 release process. This should add a conditional step in the `omero-build` role to install `exe4j` license from a string.


joshmoore:
Running against spider-5 as a final check before merging.

joshmoore:
Runs safely without the variable set (i.e. skipped).

joshmoore:
Merged manually due to travis failure which was unrelated to this PR.

PR #135, opened 2016-10-03
Ubuntu RDP

aleksandra-tarkowska:
This PR allows to adding a graphical interface to your Ubuntu Openstack instance based on https://github.com/naturalis/openstack-docs/wiki/Howto:-Adding-a-graphical-interface-to-your-Ubuntu-Openstack-instance

```
$ ansible-playbook os-rdp.yml -e vm_name=ola-test-rdp -e vm_key_name=ola

PLAY [localhost] ***************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [fail] ********************************************************************
skipping: [localhost]

TASK [fail] ********************************************************************
skipping: [localhost]

TASK [RDP VM] ******************************************************************
changed: [localhost]

TASK [RDP VM hosts files] ******************************************************
changed: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "IPs (Docker) private:192... floating:10.0..."
}

localhost                  : ok=4    changed=2    unreachable=0    failed=0   

$ ansible-playbook -l ola-test-rdp -u ubuntu desktop-rdp.yml 
Enter password: 
confirm Enter password: 

PLAY [desktop-rdp] *************************************************************

TASK [setup] *******************************************************************
ok: [371e32a6-a860-4140-bfb5-4307d3c5a473]

TASK [ubuntu-rdp : Change ubuntu password] *************************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473]

TASK [ubuntu-rdp : Get aptitude for upgrade] ***********************************
ok: [371e32a6-a860-4140-bfb5-4307d3c5a473]

TASK [ubuntu-rdp : Safe system upgrade via aptitude] ***************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473]

TASK [ubuntu-rdp : Install ubuntu-desktop] *************************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473]

TASK [ubuntu-rdp : Install base packages] **************************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473] => (item=[u'firefox', u'xrdp', u'unity-lens-applications', u'unity-lens-files'])

TASK [ubuntu-rdp : Allow using RDP] ********************************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473] => (item=[u'xfce4', u'xfce4-terminal', u'gnome-icon-theme-full', u'tango-icon-theme'])

TASK [ubuntu-rdp : Create .xsession] *******************************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473]

TASK [ubuntu-rdp : Set xfce4-session] ******************************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473]

RUNNING HANDLER [ubuntu-rdp : restart display manager] *************************
changed: [371e32a6-a860-4140-bfb5-4307d3c5a473]

371e32a6-a860-4140-bfb5-4307d3c5a473 : ok=10   changed=8    unreachable=0    failed=0   

```


ansible/roles/ubuntu-rdp/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    mode: 0644
+
+- name: Set xfce4-session
+  become: yes
+  lineinfile:
+    dest: /home/{{ rdp_user }}/.xsession
+    line: "xfce4-session"
+
+# https://support.ansible.com/hc/en-us/articles/201958037-Reboot-a-server-and-wait-for-it-to-come-back
+- name: system | reboot
```
aleksandra-tarkowska: that is copy and paste from other role, maybe should become actually a role

joshmoore:
Looks like you will need the fix from http://stackoverflow.com/questions/39829473/cryptography-assertionerror-sorry-but-this-version-only-supports-100-named-gro as well.

aleksandra-tarkowska:
yes, I will come back to this PR soon

ansible/os-rdp.yml
```diff
@@ -0,0 +1,64 @@
...
+- hosts: localhost
+  connection: local
+  #gather_facts: false
+
+  vars:
+
+    vm_image: "Ubuntu 14.04.4 LTS (trusty)"
+    vm_flavour: m1.medium
+    vm_size: 50
+    vm_groups: "ansible-managed,os-image-ubuntu-gui"
```
joshmoore: Include `ubuntu-rdp` here.

aleksandra-tarkowska:
The only thing I dont know how to handle in a playbook is user and password. Do we want give user full sudo rights or not? Should we create new user with password or reset password for default `ubuntu`?

aleksandra-tarkowska:
@mporter-gre is testing RDP now and looks like copy/paste doesn't work, I may improve that as well

joshmoore:
> The only thing I don't know how to handle in a playbook is user and password. Do we want give user full sudo rights or not? Should we create new user with password or reset password for default ubuntu?

Since the playbook asked for a password, I assumed you intended the password reset to happen. (I had to manually re-do it). For me, I think having the base playbook use the `ubuntu` (with sudo) account with a better password makes a lot of sense. If someone _then_ wants to add more users in another playbook, that's their business.

aleksandra-tarkowska:
I am afraid password reset is not quit ready yet. I wasn't sure how to reset password using user module, example only shows how to create user with password. I will get back to that soon.

ansible/os-rdp.yml
aleksandra-tarkowska: instead of `os-image-ubuntu-gui` ?

aleksandra-tarkowska:
ok password is fixed. Although I am still not sure how to make reset password idempotent

ansible/roles/ubuntu-rdp/tasks/main.yml
rleigh-codelibre: What's the reason for rebooting here?  Can't you just start/restart the necessary service?

ansible/os-rdp.yml
joshmoore: Well, your playbook as "hosts: ubuntu-rdp" here: https://github.com/openmicroscopy/infrastructure/pull/135/files#diff-a325b563c4984f92c941f54ec31972eeR4

ansible/roles/ubuntu-rdp/tasks/main.yml
aleksandra-tarkowska: simplicity, do you know what exactly would have to be restart?

aleksandra-tarkowska: simplicity, do you know what exactly would have to be restart?

rleigh-codelibre: The display manager would be the main one I can think of -- lightdm or sddm most likely.  But if there's more than that, rebooting doesn't hurt.

manics: https://github.com/openmicroscopy/infrastructure/blob/master/ansible/roles/upgrade-distpackages/tasks/main.yml#L32

Do you want to refactor that instead of duplicating? Either into a separate role, or (perhaps better?) make a pseudo-role `roles/common` for useful files, and `include: ../common/tasks/reboot.yml`? This could also be a useful place to put common handlers which are called from more than one role.

joshmoore:
Revert https://github.com/openmicroscopy/infrastructure/pull/135/commits/9a0ac90f550a14e08a250047d70cfe6a1169f1cb locally for using this at the moment.

aleksandra-tarkowska:
@joshmoore no need to revert anything https://github.com/openmicroscopy/infrastructure/pull/135#issuecomment-257844064. @rleigh-codelibre was right we just need to restart display manager. Good to test as it is

ansible/roles/ubuntu-rdp/tasks/main.yml
aleksandra-tarkowska: no, restarting display manager works :)

mtbc:
(So with X11 forwarding would this help @gusferguson get screenshots from a fairly vanilla Ubuntu system?)

joshmoore:
In general, yes. From my 10.10 laptop, I used RDC with the UoD Gateway to connect and have a working desktop. Admittedly, the one present is fairly minimal however.

gusferguson:
@mtbc - thanks for the thought - I have managed to photoshop a mockup that should do. Cheers!

manics:
Briefly discussed offline: Can you figure out why the system wants to reboot? For instance ff it's because of a general system update (e.g. the kernel) as opposed to something you've specifically done in this role then rebooting shouldn't be part of this.

Also could you add a README, especially to explain the variables?

aleksandra-tarkowska:
after wiping out virtualenv and installing latest ansible<2.2 it is no longer asking to reboot. I don't know. Try as it is and see.

aleksandra-tarkowska:
actually if I am enough patient it is there microsoft remote desktopscreensnapz001

manics:
Is that just an auto-updater rather than ansible (in which case it could potentially pop up at any point after the VM is started)?

aleksandra-tarkowska:
@manics but we upgrade packages in https://github.com/openmicroscopy/infrastructure/pull/135/files#diff-9446c54acf9368f3c038c122e06cfd4eR16

manics:
Why? Ideally the role should have one purpose (setup RDP), can you handle upgrades in a separate role, or in the playbook only? Otherwise this can't be used in an idempotent manner

aleksandra-tarkowska:
I dont have time for that

PR #136, opened 2016-10-04
Add a lookup plugin for `pass`

joshmoore:
Copied from https://github.com/gcoop-libre/ansible-lookup-plugin-pass,
this lookup plugin permits reading encrypted values using https://www.passwordstore.org

By settings ANSIBLE_PASS_PASSWORD_STORE_DIR globally, a non-standard
credentials directory can be chosen.

```

---
- hosts: localhost
  connection: local
  gather_facts: False
  tasks:
    - debug: msg="{{ lookup('pass', item) }}"
      with_items:
        - "ci/snoopy/sshconfig"
```

Also looked at:
- https://github.com/Dissonant-Tech/ansible-lookup-pass
- https://github.com/morphje/ansible_pass_lookup
- https://github.com/lbischof/ansible-role-pass

### DIFF
--- a/ansible/.gitignore
+++ b/ansible/.gitignore
@@ -1,3 +1,5 @@
 .vagrant
 *.retry
 *.un~
+.cloud*.lock
+*.pyc
--- a/ansible/lookup_plugins/pass.py
+++ b/ansible/lookup_plugins/pass.py
@@ -5,11 +5,12 @@
 # To change the default password store set ANSIBLE_PASS_PASSWORD_STORE_DIR
 # environment variable.
 #
-# If the pass doesn't exist in the store it's generated. It accepts two extra
-# parameters: length and symbols (if symbols is True or yes -n is appended to
-# the pass generate command).
+# It accepts two extra parameters: length and symbols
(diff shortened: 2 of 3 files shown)


joshmoore:
For discussion, primary issue with this seems to be making the jenkins role ever less re-usable. Considering the difficult of setting up `snoopy_dir_path`, though, I consider this an improvement.

joshmoore:
Note: this needs a modification to prevent the generation of passwords unless a flag is set.

joshmoore:
From discussion:
 * add default mechanism from https://github.com/openmicroscopy/infrastructure/pull/128
 * add support for dir back
 * move to variables so they can be overwritten
 * Auto-generating a key doesn't seem useful.

manics:
http://docs.ansible.com/ansible/playbooks_lookups.html#the-passwordstore-lookup :laughing:

joshmoore:
Very cool. Happy to update this as soon as we move to ansible 2.3.

joshmoore:
Closing this in favor of using Ansible 2.3's `passwordstore` support. Currently testing it as a replacement for `vault` then we can come back to removing `snoopy_dir_path`.

PR #137, opened 2016-10-04
omero-build: add defaults to fix spacewalk issue

joshmoore:
This works around an issue that has appeared a few times recently. I also failed to properly set this from the command-line with `-e spacewalk=True` (i.e. my tasks were still not activating)

Likely a rethink is needed of the spacewalk boolean. Perhaps it can be wrapped as an entire role.

### DIFF
--- a/ansible/roles/omero-build/defaults/main.yml
+++ b/ansible/roles/omero-build/defaults/main.yml
@@ -3,3 +3,5 @@
 
 # A license key (default null)
 exe4j_license_key:
+
+spacewalk: False


manics:
From @kennethgillen: <:ghost:>  (a fresh PXE, spacewalk-provisioned, had no EPEL definition in /etc/yum.repos.d so needed logic to make it work. the package manager could see and use yum install from EPEL, but Ansible didn't know that ) </:ghost:>

IIRC it's just a matter of waiting for spacewalk to sync with the server, so if we mostly using this for VMs and very rarely provisioning LSC servers from scratch how about accept the wait and remove this boolean completely?

joshmoore:
Merging this as a bug fix.  See https://github.com/openmicroscopy/infrastructure/issues/138 for the follow on.

PR #140, opened 2016-10-06
Handle websockets and allow streaming in nginx proxy

manics:
Add variables and template config to nginx-proxy role for proxying websockets (required by jupyterhub)


joshmoore:
Reading along with the recent work on jupyter & rsync, this makes sense to me.

PR #114, opened 2016-09-13
devspace playbook

aleksandra-tarkowska:
This PR shows how to deploy and manage devspace using ansible and docker_service

## Minimum requirements:
 - brief understanding of ansible http://docs.ansible.com/ansible/intro_getting_started.html or tutorial by @joshmoore 
  - inventory http://docs.ansible.com/ansible/intro_inventory.html
  - playbook http://docs.ansible.com/ansible/playbooks.html
 - access to openstack tenancy
 - own ssh key set in openstack tenancy, that name will be used as `vm_key_name`
 - `openrc.sh` http://docs.openstack.org/user-guide/common/cli-set-environment-variables-using-openstack-rc.html
 - snoopy ssh key and gitconfig (* to be discussed

## Deployment
To run devspace run following on the client machine:

- install virtualenv and pip prerequisites:

 ```
$ virtualenv dev
$ source dev/bin/activate
(dev) $ pip install ansible
(dev) $ pip install shade

(dev) $ git clone -b devspace https://github.com/aleksandra-tarkowska/infrastructure.git
(dev) $ cd infrastracture/ansible
```

- create vm (**boot from volume, you no longer have to attach additional volumes**)

 ```
(dev) $ source path/to/openrc.sh
# vm_key_name is a name of ssh key in openstack
# vm_size (default 50GB) is a size of the volume vm boot from. You no longer have to attach additional volumes!
(dev) $ ansible-playbook os-devspace.yml -e vm_name=my-devspace -e vm_key_name=mysshkey
```

  **write down IP of newly created vm, you will need it below to set `openstack_ip`**

- create inventory:

 ```
$ tree /path/to/inventory
devspace
      ├── devspace-hosts
      ├── group_vars
      │   └── devspace
      └── snoopy
          ├── .gitconfig
          └── .ssh

 ```

 - `/path/to/inventory/devspace/devspace-hosts`

    ```
    [devspace]
    10.0.50.100
    ```

 - put snoopy .ssh in `/path/to/inventory/devspace/snoopy/.ssh` that includes:

    ```
    -rwx------.  1    74 Sep 13 15:25 config
    -rwx------.  1  1674 Sep 13 15:25 snoopycrimecop_github
    -rwx------.  1   405 Sep 13 15:25 snoopycrimecop_github.pub
    ```

 - put snoopy gitconfig in `/path/to/inventory/devspace/snoopy/.gitconfig`

 - `/path/to/inventory/devspace/group_vars/devspace`

    ```
[...]

PR #141, opened 2016-10-08
Devspace role unification

sbesson:
Following a recent conversation with @manics and @aleksandra-tarkowska, this PR refactors the 2 devspace deployment playbooks introduced in https://github.com/openmicroscopy/infrastructure/pull/114 as sub-tasks of the devspace.

This allows both tasks to share the common variables and have only 1 call to install and then start the containers. From this branch I was able to update the BF 5.3 devspace with:

```
$ ansible-playbook devspace.yml -u centos -l seb-datarepo
```

where my only local addition was a `seb-datarepo` host var file with

```
docker_use_ipv4_nic_mtu: True
git_repo: "https://github.com/sbesson/devspace.git"
version: "bf_devspace"
omero_branch: dev_5_3
```

This PR also pins the devspace branch to the last tag (we can bump it via PR - makes the role more stable) and uses `force/update` while cloning the devspace (i.e. `git reset --hard`).

--

With the latest state of this branch, I have an issue on the BF devspace where HTTPS URLS can no longer be cloned inside the docker containers and is hanging

```
[omero@seb-datarepo ~]$  git clone  https://github.com/openmicroscopy/bioformats.git
Cloning into 'bioformats'...
remote: Counting objects: 241951, done.
remote: Compressing objects: 100% (32/32), done.
^Cceiving objects:   9% (21776/241951), 1.77 MiB | 3.35 MiB/s   
[omero@seb-datarepo ~]$ docker exec -it devspace_testintegration_1 bash

bash-4.2$ GIT_CURL_VERBOSE=1 GIT_TRACE=1 git clone https://github.com/openmicroscopy/bioformats.git /tmp/bf
trace: built-in: git 'clone' 'https://github.com/openmicroscopy/bioformats.git' '/tmp/bf'
Cloning into '/tmp/bf'...
trace: run_command: 'git-remote-https' 'origin' 'https://github.com/openmicroscopy/bioformats.git'
* Couldn't find host github.com in the .netrc file; using defaults
* About to connect() to github.com port 443 (#0)
*   Trying 192.30.253.112...
* Connected to github.com (192.30.253.112) port 443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
^C
bash-4.2
```

EDIT: found the origin of the issue above `docker_use_ipv4_nic_mtu: True` needs to be overriden in the host vars as it is set to False by default in the docker role.


aleksandra-tarkowska:
network issue is knows and was pointed to set `docker_use_ipv4_nic_mtu: True` in https://github.com/openmicroscopy/infrastructure/pull/114. @manics is there any reason why default is False?

sbesson:
While working on fixing the networking issue discussed above, I found a secondary issue in the current design of this playbook. If the docker options are modified e.g. by changing `docker_use_ipv4_nic_mtu`, then the playbook terminates with the execution of the `docker restart` handler from the `docker` role which stops the created devspace.

aleksandra-tarkowska:
is that because include is not static?

sbesson:
Adding `include` globally didn't solve the problem after making a change to the `docker` configuration options. This matches what I expect from the [Ansible documentation](http://docs.ansible.com/ansible/playbooks_intro.html#handlers-running-operations-on-change) though:

```
Roles are described later on, but it’s worthwhile to point out that:

    handlers notified within pre_tasks, tasks, and post_tasks sections are automatically flushed in the end of section where they were notified;
    handlers notified within roles section are automatically flushed in the end of tasks section, but before any tasks handlers.
```

So the various solutions I can think of are that `devspace-runtime` could be executed as a post_tasks, as a role handler (assuming handlers are flushed out in the order of the roles) or back to a separate playbook...

sbesson:
Last commit should migrate the restart logic as a role handler. Tested against a running devspace. Also checked the logic when also changing both `docker` option in check-mode:

```
(ansible) sbesson@ls30630:ansible (devspace_role_unification) $ ansible-playbook -u centos devspace.yml -l seb-datarepo -C
statically included: /opt/ome/management_tools/infrastructure/ansible/roles/devspace/tasks/devspace-install.yml
statically included: /opt/ome/management_tools/infrastructure/ansible/roles/devspace/tasks/devspace-runtime.yml
Enter Jenkins password: 
...
TASK [devspace : rename to dev_5_3...] *****************************************
skipping: [seb-datarepo]

RUNNING HANDLER [docker : restart docker] **************************************
changed: [seb-datarepo]

RUNNING HANDLER [devspace : Devspace down] *************************************
changed: [seb-datarepo]

RUNNING HANDLER [devspace : Devspace up] ***************************************
changed: [seb-datarepo]

RUNNING HANDLER [devspace : debug] *********************************************
ok: [seb-datarepo] => {
...
RUNNING HANDLER [devspace : assert] ********************************************
ok: [seb-datarepo]

seb-datarepo               : ok=27   changed=7    unreachable=0    failed=0   

(ansible) sbesson@ls30630:ansible (devspace_role_unification) $ 
```

ansible/roles/devspace/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+- name: restart devspace
+  environment:
+    USER_ID: "{{ user_id }}"
+  become: yes
+  become_user: "{{ devuser }}"
+  include: devspace-runtime.yml
```
aleksandra-tarkowska: Do you think name of the file could become a var? I have more complex playbook as an example of managing containers https://github.com/aleksandra-tarkowska/infrastructure/blob/extend_devspace/ansible/devspace-extend.yml

joshmoore:
No vetoes. Merging. Follow-up is to rename all variables with `devspace_` prefix.

joshmoore:
Note: c8291166f90f02598bb82be56898a2b1d63bffd4 surprised me and led to dataloss on idr-ci. For me, this makes this a non-deploy (i.e. a provisioning) playbook.

PR #148, opened 2016-10-17
Work around websocket proxy issue: set Host header

manics:
Follow up to https://github.com/openmicroscopy/infrastructure/pull/140. Adds a workaround for webproxy connections where the `Host` header seems to be incorrect, even though it's already set in the `server` scope.


sbesson:
Discussed at meeting, this has been deployed. Merging.

PR #152, opened 2016-10-25
New role to hide SELinux issues

manics:
There are several roles which require conditional tasks depending on whether SELinux is enabled or not. Furthermore Ansible has some SELinux package dependencies which aren't always installed by default. This role abstracts away all that complexity:
- Sets a host-variable `selinux_enabled`: `True`/`False`, this can be checked in other roles or playbooks
- If SELinux is enabled or permissive installs SELinux dependencies required by Ansible, otherwise it doesn't
- See the README for details.

This also modifies the `nginx-proxy` role to use this new role. Other roles can be updated in separate PRs.

### DIFF
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -51,6 +51,13 @@ Vagrant.configure(2) do |config|
     end
   end
 
+  config.vm.define "selinux" do |selinux|
+    selinux.vm.box = "centos/7"
+    selinux.vm.provision "ansible" do |ansible|
+      ansible.playbook = "test.yml"
+    end
+  end
+
   config.vm.define "training-server" do |training|
     training.vm.box = "centos/7"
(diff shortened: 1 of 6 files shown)


joshmoore:
```
PLAY [selinux] *****************************************************************

TASK [setup] *******************************************************************
ok: [selinux]

TASK [selinux-utils : selinux | check enabled] *********************************
ok: [selinux]

TASK [selinux-utils : system packages | set selinux variable] ******************
ok: [selinux]

TASK [selinux-utils : system packages | install selinux utilities] *************
changed: [selinux] => (item=[u'libselinux-python', u'libsemanage-python', u'policycoreutils-python'])

TASK [assert] ******************************************************************
ok: [selinux]

selinux                    : ok=5    changed=1    unreachable=0    failed=0
```

LGTM.

PR #153, opened 2016-10-25
Use docker swarm with IDR JupyterHub

manics:
This updates the IDR docker playbook to run containers on Docker Swarm. At the moment this requires custom changes to
- Docker Spawner: https://github.com/jupyterhub/dockerspawner/compare/master...manics:master
  Docker 1.12 only supports Swarm mode for services (Note Docker 1.13 should support running normal containers in the swarm https://github.com/docker/docker/pull/25962 https://github.com/jupyterhub/dockerspawner/issues/117#issuecomment-255141210)
- Several networking MTU workarounds
- Also note that the main JupyterHub server is run as a Docker service, which potentially conflicts with Systemd since Docker service will automatically start/restart it

### DIFF
--- a/ansible/idr-playbooks/files/jupyterhub-docker-swarm.j2
+++ b/ansible/idr-playbooks/files/jupyterhub-docker-swarm.j2
@@ -1,0 +1,33 @@
+[Unit]
+Description=IDR JupyterHub
+Requires=docker.service
+After=docker.service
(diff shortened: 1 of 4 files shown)


manics:
See https://github.com/openmicroscopy/infrastructure/pull/164 instead

PR #156, opened 2016-10-26
System upgrade fixes

manics:
- Force a yum cache refresh. Today's patching showed that the cache isn't always updated before instaling updates.

### DIFF
--- a/ansible/roles/upgrade-distpackages/tasks/main.yml
+++ b/ansible/roles/upgrade-distpackages/tasks/main.yml
@@ -6,6 +6,7 @@
   yum:
     name: "{{ item }}"
     state: latest
+    update_cache: yes
   with_items: "{{ upgrade_distpackages }}"
 
 - name: system | current running kernel


manics:
This PR originally added `upgrade_distpackages_reboot_kernel: True` to `all-upgrade.yml` to automatically reboot a server if the kernel was upgraded, but I took it back out. Any thoughts on whether it should be the default in the playbook?

```
diff --git a/ansible/all-upgrade.yml b/ansible/all-upgrade.yml
index 144b400..48ce6b2 100644
--- a/ansible/all-upgrade.yml
+++ b/ansible/all-upgrade.yml
@@ -4,4 +4,3 @@
 - hosts: all
   roles:
   - role: upgrade-distpackages
+    upgrade_distpackages_reboot_kernel: True
```

joshmoore:
I don't have any opinion on the defaults. Otherwise, seems fine as is.

manics:
Anyone else have thoughts @sbesson @kennethgillen?

sbesson:
No strong opinion about `upgrade_distpackages_reboot_kernel`. If this is something we want to enable selectively depending on the host/playbook, we could reintroduce it with a variable.

joshmoore:
No vetoes.

kennethgillen:
@manics - depending on how we're driving the upgrades. safest to leave 'reboot server' out, since it's likely to be more disruptive than the actual patch. The reboots are best done in a highly controlled manner.

PR #161, opened 2016-11-03
Omero python deps

manics:
Move python requirements that are common to OMERO.server/OMERO.py and OMERO.web into a separate role. Adds configuration options for:
- recommended vs minimal python deps: in theory the minimal deps are enough for web, in practice most people should just install all the deps (default)
- options for choosing mostly distribution packages (default) vs mostly pypi packages. This currently only affects pillow but could be extended to cover most python requirements

This means the existing `omero-runtime` role (which `omero-server` depends on) is now just a "meta" role which only installs dependent roles, so could arguably be completely removed.

This is mostly aimed at runtime requirements, this doesn't touch the `python-devel` role.

`omero-web-runtime` now only installs web-specific requirements (gunicorn, django)


manics:
This fixes the IDR user photo:
screen shot 2016-11-03 at 18 13 57

aleksandra-tarkowska:
just for the future. `ansible/roles/omero-python-deps/tasks/pypi.yml ` could be fully configurable from vars. Ideally if pip/yum packages could go to {{ omero_py_pip_dependencies }} like:

```
- name: omero.py pip dependencies
  become: yes
  yum:
    name: "{{ item }}"
    state: present
  with_items: "{{ omero_py_pip_dependencies }}"
  when: omero_py_pip_dependencies

- name: omero.py pip dependencies
  become: yes
  pip:
    name: "{{ item }}"
    state: present
  with_items: "{{ omero_py_pip_packages }}"
  when: omero_py_pip_packages

```

aleksandra-tarkowska:
sorry for being picky, is zeroc-icepy not OMERO.py requirements?. How do we want to handle that? Ice role installs much more then just icepy.

manics:
Yes, that's why I added "This role does not install the Ice Python module." so we can leave that for later.

joshmoore:
* @aleksandra-tarkowska : ok to handle ice in a follow-on?
 * @manics : I assume this should result in no change for most servers?

manics:
>  I assume this should result in no change for most servers?

Recently there was a fight between two roles, one installing the `python-pillow` rpm, the other removing the rpm and using `pip install pillow`, so some servers may end up with both installed, or may already have both.

joshmoore:
Merging for deployment at EBI. Follow ups to ... follow.

PR #163, opened 2016-11-07
Download nginx from https location

manics:
I updated VirtualBox and was unable to download http://nginx.org/packages/centos/7/noarch/RPMS/nginx-release-centos-7-0.el7.ngx.noarch.rpm when running vagrant. I've no idea if it's a VirtualBox/OS X problem, or a change by nginx.org, but using `https` instead seemed to fix it. In any case using `https` should be more secure.

### DIFF
--- a/ansible/roles/nginx/tasks/main.yml
+++ b/ansible/roles/nginx/tasks/main.yml
@@ -5,7 +5,7 @@
 - name: system packages | setup upstream nginx repo
   become: yes
   yum:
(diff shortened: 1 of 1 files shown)


manics:
I've just run into another download error with `pip`, so I'm guessing it's a VirtualBox networking bug.

joshmoore:
Do you need to push another change here, or is this good to go?

manics:
It's good from my point-of-view

PR #165, opened 2016-11-08
Add openstack provider to Vagrantfile

manics:
Allow you to test vagrant boxes on openstack, useful if you upgrade VirtualBox and discover it's broken.

```
vagrant plugin install vagrant-openstack-provider
source openrc.sh
vagrant up --provider openstack BOXNAME ...
```

### DIFF
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -11,6 +11,22 @@ Vagrant.configure(2) do |config|
     vb.customize ["modifyvm", :id, "--memory", "2048"]
   end
 
+  config.vm.provider :openstack do |os|
+    config.ssh.username = "centos"
+    # Workaround sudo restriction
+    config.ssh.pty = true
+
+    os.openstack_auth_url = "#{ENV['OS_AUTH_URL']}"
+    os.username = "#{ENV['OS_USERNAME']}"
+    os.password = "#{ENV['OS_PASSWORD']}"
+    os.tenant_name = "#{ENV['OS_TENANT_NAME']}"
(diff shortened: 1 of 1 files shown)


joshmoore:
Only minor issue is probably that the name of host created matches that in Vagrantfile so is pretty non-specific. Also note: `vagrant destroy --provider=openstack ...` doesn't work.

manics:
You only pass `--provider openstack` to vagrant up, all other commands operate on existing VMs and Vagrant keeps track of the provider.

ansible/Vagrantfile
```diff
@@ -11,6 +11,22 @@ Vagrant.configure(2) do |config|
     vb.customize ["modifyvm", :id, "--memory", "2048"]
   end
 
+  config.vm.provider :openstack do |os|
+    config.ssh.username = "centos"
```
joshmoore: I think this broke non-openstack uses. Adding `node.ssh.username = "vagrant"` in the large loop corrects the issue. cc: @jburel

manics: Does adding `node.ssh.username = "vagrant"` to just the `config.vm.provider "virtualbox" do |vb|` also work?

jburel: ``config.ssh.username = "centos"`` certainly break non-openstack usage. I replaced locally ``centos`` by ``vagrant`` I will try that ``node.ssh.username`` option

joshmoore: What should `    config.vm.define "#{server}" do |node|` be in that case?

PR #166, opened 2016-11-08
OpenStack volume ansible 2.3 features

manics:
https://github.com/ansible/ansible-modules-core/pull/5176 enables Ansible to clone openstack volumes, which is quicker than creating a volume from a snapshot. It is targeted for Ansible 2.3, but in the meantime this PR bundles it in the `openstack-volume-storage` role so we can take advantage of it in the IDR. This should be reverted assuming Ansible 2.3 is released with the above feature.

It should be possible to override a module of the same name, but I couldn't get it to work.

### DIFF
--- a/ansible/roles/openstack-volume-storage/README.md
+++ b/ansible/roles/openstack-volume-storage/README.md
@@ -3,6 +3,9 @@ Openstack Volume Storage
 
 Create and attach a volume to an Openstack instance.
 
+Note this requires Ansible 2.3 features in `os_volume`.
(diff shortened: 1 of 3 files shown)


joshmoore:
Ah, nice. Didn't realize you could have a library per role. This is what I did locally to make things work (though I named it `volume_idr`... :smile:) because without it the previous role is effectively broken. Merging.

manics:
When I tested the role it only failed if you used the new parameter, otherwise it behaved as previously.

PR #168, opened 2016-11-10
Remove openstack/examples (out of date)

manics:
I think these are old scripts leftover from when we were testing OpenStack with the IDR. I can't think of a reason to keep it.


joshmoore:
Only the one might still be useful, but if/when we can grab it from the history.

PR #170, opened 2016-11-14
Remove redundancy in Vagrantfile test.yml definitions

manics:
Replace 8 identical server definitions with a loop. Removed Ansible verbose output so it's easier to check the output, there should be no other change in behaviour or usage of vagrant.


joshmoore:
Works fine. Thanks.

PR #172, opened 2016-11-15
Jekyll role

manics:
Copied the tasks for installing Jekyll and deploying a site from `jekyll-playbooks/`


joshmoore:
Any objections to updating the playbooks to use the new role? cc @sbesson

sbesson:
I was about to ask the same thing. No veto from my side if we are converting these playbooks into proper roles.

manics:
I'm doing that in a separate PR

sbesson:
I assume leaving out the nginx deployment part of the playbook was on purpose. In that case, it's unclear to me that `jekyll-site` is the right role name as this stops at the Jekyll build step. Otherwise, is there room for adding a nginx dependency with a default template? And could the latter be overriden in playbooks?

manics:
Yes, it's intentional because there's too many ways to use nginx in production, e.g. serving a single jekyll site, serving multiple, serving jekyll and proxying a backend server, so I think it should be separate. I'm happy to rename `jekyll-site` to something else if you've got a suggestion.

ansible/test.yml
```diff
@@ -95,6 +95,21 @@
       that:
       - "sphinx_version.stdout == 'Sphinx (sphinx-build) 1.2.3'"
 
+- hosts: jekyll
+  roles:
+  - role: jekyll-site
+    jekyll_site_name: test
+    jekyll_site_git_repo: https://github.com/lunson/www.openmicroscopy.org
```
sbesson: We might want to change this to use an organization repo. IDR?

manics: The proper www.openmicroscopy.org repo would make more sense, but as far as I can tell there's not much there yet.

sbesson: That would work as well, we'll have to remember to update this test file eventually.

manics:
Updated

sbesson:
All works as expected both using Vagrant and the updated playbook in #173. @joshmoore any additional comment?

joshmoore:
Not from me. :+1: Thanks, guys.

sbesson:
Thanks @manics for the consolidation.

PR #175, opened 2016-11-16
Bump default devspace version to 0.3.4

sbesson:
This should include the bug fix for handling submodules in the OMERO build /cc @aleksandra-tarkowska


joshmoore:
@aleksandra-tarkowska deploying a new devspace today

PR #176, opened 2016-11-16
remove gcc from basedeps

aleksandra-tarkowska:
include gcc in OMERO.py


manics:
Cherry-picked onto my IDR branch, looks good so far.

PR #177, opened 2016-11-17
Openstack IDR roles (IDR-0.3.0)

manics:
This contains the majority of the changes to the IDR specific roles from https://github.com/manics/infrastructure/tree/idr-0-3-0: `openstack-idr-{instance,keypair,network}`


joshmoore:
Testing while creating adhoc docker servers at UoD and EBI. Primary "post-processing" issue I had (which may just have been a matter of my not seeing the documentation) is how to add adhoc groups.

Otherwise, network, router, keys, instance all worked fine. Cheers. Merging.

PR #178, opened 2016-11-17
Network basic interface addition (IDR-0.3.0)

manics:
- Adds a tasks for adding a single NIC without any fancy configuration
  - This was required because a full network restart was causing problems on one of the IDR instances. This addition should only be used where unavoidable.
- Allows the MTU to be unset (i.e. managed by the system).

### DIFF
--- a/ansible/roles/network/README.md
+++ b/ansible/roles/network/README.md
@@ -13,6 +13,13 @@ Role Variables
 - `network_disable_ifaces`: A list of network device names to be explicitly disabled, use this if you want to be sure the interface is disabled (as opposed to being auto-configured by the system).
 - `network_delete_ifaces`: A regular expression describing the network device name(s) to be removed (note this means the system may auto-configure them), use this for cleaning up spare configuration files.
 
+The above variables provide full control of all network interfaces on a server, and will almost always require a restart of the network service on the server.
+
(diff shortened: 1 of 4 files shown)


manics:
The alternative is to move the new file into a separate role. It means some duplication, but might keep things cleaner?

joshmoore:
Think that's beyond the complexity of networking that I've tried so far. The state of this PR certainly worked well for my doing simpler "new network+router"-style deployments for the adhoc docker services (gh-177). Happy for this to be merged as is, or for you to refactor as needed. Leaving the merging to you.

manics:
Thanks, I'm going to hold off merging this until we tag idr-0.3.0, since this is will only be required for idr-0.3.1.

manics:
I was wrong, this is required by https://github.com/openmicroscopy/infrastructure/pull/192 since the EBI server has two NICs.

manics:
This should only be needed with VMs, and I've figured out a simpler way to do this with a new role. Options for this PR are to merge since it's required by the IDR 0.3.0 playbooks and revert some of the commits in a future PR, or to close this and (temporarily) loose the ability to rebuild the IDR at EBI from scratch (UoD is fine because there's only a single network).

joshmoore:
I would very, _very_ much like to be able to re-do 0.3.0 while you are away since that was exactly my problem recently. Merging and we can revert as necessary.

PR #179, opened 2016-11-17
Openstack dynamic inventory: store and detect SSH proxy (IDR-0.3.0)

manics:
Use optional metadata properties on OpenStack instances to:
1. Automatically set an SSH proxy host
   - If the env-var `OS_PROXY_DISCOVER=1` and property `ssh_proxy_host` is set on the instance:
     - `ssh_proxy_host=proxy` indicates this instance should be used as a SSH proxy for any instances connected to the same network
     - `ssh_proxy_host=required` means attempt to find an instance on the same network with `ssh_proxy_host=proxy` and if found set the host-variable `ansible_ssh_common_args` with `ProxyCommand`, unless that hostvar already has a value.
   - The env-var is required to avoid surprising people who use this dynamic inventory without being aware of the extra features.
2. Return networks in the original order
   - The OpenStack dynamic inventory returns the list of networks that an instance is attached to as a dictionary, so it is not possible to retrieve the order that the networks were defined as at creation time. If the property `network_order` is set then this must be a comma separated list of network names in order.

Use in conjunction with changes to `roles/openstack-idr-instances` in https://github.com/openmicroscopy/infrastructure/pull/177

### DIFF
--- a/ansible/inventory/openstack-private.py
+++ b/ansible/inventory/openstack-private.py
@@ -132,6 +132,51 @@ def append_hostvars(hostvars, groups, key, server, namegroup=False):
         groups[group].append(key)
 
 
+def is_ssh_proxy_host(server, network):
+    """
+    Checks the metadata to see if this is an SSH proxy host.
+    If it also has a floating IP then return this, otherwise return None.
+    """
+    try:
+        if not server['openstack']['metadata']['is_ssh_proxy_host']:
+            return
+    except KeyError:
+        return
+
+    for port in server['openstack']['addresses'][network]:
+        if port['OS-EXT-IPS:type'] == 'floating':
+            return port['addr']
+
+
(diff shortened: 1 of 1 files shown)


joshmoore:
Initial testing (using `-i inventory/openstack-private.py` for provisioning) worked well and certainly fixed the issue I was having. I *didn't* test *any* of the other possible workflows.

joshmoore:
@manics : you mentioned something not throwing that you expected to throw. How would you like to handle this with regard to this PR?

manics:
I think that was due to an Ansible ssh related variable which was defined in a group-vars file (I subsequently removed it in another PR)

joshmoore:
Kk. Merging as a general improvement, but we'll still all need to be a bit careful about running against multiple clouds, etc.

PR #158, opened 2016-10-27
Sphinx build role

sbesson:
This is a minimal port of the changes proposed https://github.com/openmicroscopy/infrastructure/pull/20 in the context of the upcoming Bio-Formats 5.3.0-m1 release preparation.

This PR puts the minimal requirements into place for documentation nodes, i.e. installing Sphinx as a dependency. As discussed previously, Sphinx is installed globally instead of within a virtual environment to allow it to be consumed by the Jenkins Ant targets which cannot activate a virtual environment. It also applies the policy of dropping the PDF in favor of ZIPs of the HTML and removes the `texlive` step from the `ci-deployment` playbook. /cc @jburel

PR #183, opened 2016-11-18
Virtualenv

aleksandra-tarkowska:
This PR removes PyPI dependences to virtualenv

This PR is for the discussion on #sysadmin and maybe some clean-up first

cc: @sbesson @manics @jburel @joshmoore


ansible/ci-deployment.yml
```diff
@@ -4,6 +4,20 @@
 - hosts: ci-jenkins-linux
   roles:
     - role: jenkinsslave
+    - role: virtualenv
+      virtualenv_user: "{{ jenkinsuser }}"
+      virtualenv_path: /home/{{ jenkinsuser }}/virtualenv
+      virtualenv_site_packages: "yes"
+      virtualenv_pip_rpm_dependencies:
+      virtualenv_pip_dependencies:
```
aleksandra-tarkowska: what to install by default in any node?

ansible/ci-deployment.yml
```diff
@@ -33,8 +47,26 @@
...
+    - role: virtualenv
     - role: sphinx-build
     - role: ansible
+
+  vars:
+    virtualenv_user: "{{ jenkinsuser }}"
+    virtualenv_path: /home/{{ jenkinsuser }}/virtualenv
+    virtualenv_site_packages: "yes"
+    virtualenv_pip_rpm_dependencies:
+    virtualenv_pip_dependencies:
```
aleksandra-tarkowska: doc deps?

ansible/ci-deployment.yml
sbesson: Should this be a variable that can be overriden by group?

ansible/roles/omero-python-deps/README.md
```diff
@@ -28,9 +28,28 @@ Role Variables
 --------------
 
 Optional variables:
-- `omero_python_deps_profile`: Either `distribution` to use the distribution packages for Python modules where possible (default), or `pypi` to install recommended packages globally from PyPI using `pip`.
 - `omero_python_deps_recommended`: If `False` install a minimal set of Python dependencies, default `True`.
 
+If `omero_python_deps_recommended` False then use virtualenv role
```
aleksandra-tarkowska: @manics are you ok with removal of omero_python_deps_profile and handling that as described below?

ansible/ci-deployment.yml
sbesson: I am not sure of the requirement for a virtualenv in ci-docs groups. As discussed in https://github.com/openmicroscopy/infrastructure/pull/158, the current nature of the documentation build stack (i.e. Ant + Sphinx) makes it not usable with a virtualenv in a Jenkins environment. Is there another use case?

aleksandra-tarkowska: @sbesson for autogen I had to add https://github.com/openmicroscopy/infrastructure/pull/129. Does it mean is no longer the case? We don't need venv at all?

sbesson: Briefly discussed with @aleksandra-tarkowska, I think the issue is that different jobs have different strategies. Some of them use Jenkins Ant + globally installed Sphinx (BIOFORMATS build jobs) others use globally installed Ant + virtualenv  (OMERO docs autogen jobs). Rationalizing might be a long term goal. To maintain compatibility, my feeling is that installing Sphinx in the system packages via `sphinx-build` and creating a virtualenv with `--system-site-packages` should maintain compatibility for both types of documentation jobs.

ansible/roles/omero-python-deps/tasks/main.yml
```diff
@@ -1,12 +1,15 @@
 ---
 # tasks file for roles/omero-runtime
 
-- name: omero python deps | check variables
-  fail:
-    msg: "Invalid value of omero_python_deps_profile: {{ omero_python_deps_profile }}"
-  when: "omero_python_deps_profile not in ['distribution', 'pypi']"
 
-# TODO: Consider installing some of these with pip if the user requests it?
+- name: omero python deps | pypi pillow uninstall
```
manics: Can you avoid uninstalling/removing packages:
- If other packages depend on this it may break them (e.g. in a previous PR removing `python-pillow` also removed `cloud-init`)
- In this particular case the global `pip` will also remove the files installed by yum, but the yum/rpm database still lists the package as installed, e.g.

```
[root@5a85874b4943 /]# yum install -q -y epel-release

[root@5a85874b4943 /]# yum install -q -y python-pip

[root@5a85874b4943 /]# pip freeze |grep Pillow

[root@5a85874b4943 /]# yum install -q -y python-pillow

[root@5a85874b4943 /]# pip freeze |grep Pillow
Pillow==2.0.0

[root@5a85874b4943 /]# python -c 'import PIL'

[root@5a85874b4943 /]# pip uninstall -y Pillow
Uninstalling Pillow-2.0.0:
  Successfully uninstalled Pillow-2.0.0

[root@5a85874b4943 /]# pip freeze |grep Pillow

[root@5a85874b4943 /]# rpm -q python-pillow
python-pillow-2.0.0-19.gitd1c6db8.el7.x86_64

[root@5a85874b4943 /]# python -c 'import PIL'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named PIL
```

ansible/roles/virtualenv/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: virtualenv | pip compilation requirements {{ virtualenv_pip_rpm_dependencies }}
+  become: yes
+  yum:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ virtualenv_pip_rpm_dependencies }}"
+
+- name: virtualenv | pypi {{ virtualenv_pip_dependencies }}
+  become: yes
+  become_user: "{{ virtualenv_user }}"
```
manics: Can you make `virtualenv_user` default to `root`, since that's what you'd probably use in production.

ansible/roles/omero-python-deps/README.md
manics: In principle that's fine. What's the plan for specifying the list of pip packages- will the full list always have to be specified in the playbook?

manics:
A more general question: Is this role intended to cover all uses of virtualenv, including use-cases which aren't related to OMERO/BF and related applications?

aleksandra-tarkowska:
@manics could you give the example of what you are thinking of?

manics:
Suppose I want to setup a dedicated box for running infrastructure tasks (either in jenkins, rundeck, or some other manager). This would require ansible, shade, python-openstackclients, and maybe some other python modules, but this has nothing to do with OMERO.

If you want this role to be focussed on OMERO you can setup things by default, if it's meant to work for every use case that's more difficult.

aleksandra-tarkowska:
I am not sure I understand why setting 
```
virtualenv_pip_dependencies:
 - ansible
 - shade
```
won't work in that case?

maybe I will add a quick test...

manics:
It should work. My question was about what your intentions are. If you keep this completely generic you'll have to bear that in mind if you need to make future changes to the role, since it could be used anywhere. If you limit the scope you've got more freedom.

PR #184, opened 2016-11-22
IDR docker (IDR-0.3.0)

manics:
This is all the commits related to IDR JupyterHub, including the history before this was refactored into the `idr-jupyter` role, but not including the last commit setting the deployed tagged versions.


joshmoore:
Changes seem ok (though more than can be fully reviewed). 2 questions:

 * The tag version bump will be in a follow up?
 * What's the impact of subnet configurability TODO?

manics:
Tags are in https://github.com/manics/infrastructure/commit/4668d16b3158fcc80fde9209fb9729689b1b18f8

> What's the impact of subnet configurability TODO?

I don't know. I copied the subnet from the `docker_gwbridge` network automatically created by Docker swarm. Ansible 2.2 has a [docker_network](https://docs.ansible.com/ansible/docker_network_module.html) which might (or might not) help.

joshmoore:
Ok. Think we have some upcoming ansible/docker network arm-wrestling, but getting this in for 0.3.0. Cheers.

PR #185, opened 2016-11-22
Remove old /about files from idr-playbooks/files/html (IDR-0.3.0)

manics:
These are now in https://github.com/IDR/idr-demo.openmicroscopy.org


joshmoore:
:+1: Thanks.

PR #186, opened 2016-11-22
Name fields in `omero.web.ui.top_links` (omero-web-apps role) (IDR-0.3.0)

manics:
The tuples that go into `omero.web.ui.top_links` are a bit of a mystery at first. This PR modifies the `omero-web-apps` role so that each of the tuple elements is named.

These field names are based on the code in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/decorators.py#L162 (I couldn't find any other docs)


aleksandra-tarkowska:
looks good

aleksandra-tarkowska:
just a general comment, not related to this PR, do you think would be worth linking omero properties to glossary page?

manics:
If we expect people to come up with their own (e.g. setting up a public mapr with customised multiple tabs) I think linking to the relevant Django properties would be useful.

PR #190, opened 2016-11-22
The IDR ansible command helper (IDR-0.3.0)

manics:
Figures out the ansible-playbook command to run when deploying the IDR for the first time.


joshmoore:
This is an interesting strategy. Happy to have merged as is since I've certainly tested successfully. Longer-term thoughts:

 * allow configurable (random?) file name generation
 * allow injecting further playbooks (and/or globbing)
 * double-check to what extent this

manics:
The main reason for this was to automatically set the ssh `ProxyCommand`. That's now handled by the `openstack-pyivate.py` dynamic inventory (https://github.com/openmicroscopy/infrastructure/pull/179).

joshmoore:
i.e. we may be moving away from it. WFM.

PR #192, opened 2016-11-22
Fix the EBI NFS playbook (IDR-0.3.0)

manics:
This includes adding `nolock` to the NFS mount options (not sure why it's required, but it failed without it), and using a modified network role to setup the second NIC.


joshmoore:
Can't see anything to object to...

PR #195, opened 2016-11-23
Add option to reuse jupyterhub container (IDR-0.3.0)

manics:
The original jupyterhub setup (with github auth) recreated the containers on every startup, since all persistent data was stored on a volume.

The temporary system auth setup requires a workaround to ensure the container is reused, since the system user accounts must be manually created. This means if you want to update the container you must manually delete it.


joshmoore:
As an aside, we mentioned recently that this PR would be during review, however:

|     When    |      State      |  How? |
|----------|:-------------:|------:|
| During review |  OS login | this PR |
| Post-publication |    githubauth   |   revert PR |
| Pre-print | Warn by email? |    ???  |

Regardless, this is the state of 0.3.0 in production so merging.

PR #196, opened 2016-11-23
Add Vagrantfile for IDR (IDR-0.3.0)

manics:
This is an attempt to get the entire IDR tested with Vagrant.

### DIFF
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -75,36 +75,122 @@ Vagrant.configure(2) do |config|
     end
   end
 
-  # IDR multinode omero
-  ["idr-database", "idr-omero", "idr-gateway"].each do |s|
-    config.vm.define "#{s}" do |node|
-      node.vm.box = "centos/7"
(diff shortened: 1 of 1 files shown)


joshmoore:
* I tested this successfully up to the point where `idr-omero` wanted to talk to `idr-database`. Both have the same `10.0.2.*` IPs for eth0 but differing for eth1 in the `172.28.128.*` range. I assume this will require similar to the ansible network-order fix
 * The order of the `vagrant up` arguments seems to also be important: `idr-database idr-omero ...`

manics:
> I assume this will require similar to the ansible network-order fix

It's a different issue. The default vagrant network (which supports external access) doesn't allow inter-vm networking, so you need a second one. I think there's typo somewhere.

> The order of the vagrant up arguments seems to also be important

That's a feature of vagrant with multiple-VMs. It brings up one VM, run ansible, then brings up the next...

manics:
The relevant playbooks are being migrated to https://github.com/IDR/deployment/

PR #197, opened 2016-11-23
Remove non-specific management-hosts from idr-local-users.yml (IDR-0.3.0)

manics:
Minor cleanup, the management VM is in `{{ idr_environment }}-hosts` so doesn't need to be specified separately.


joshmoore:
LGTM

PR #200, opened 2016-11-25
Automatically initialise multiple NICs in an centos-7 cloud instance (IDR-0.3.1)

manics:
This role will go through all network interfaces detected by Ansible that don't have a corresponding `/etc/sysconfig/network-scripts/ifcfg-*` file, and will configure that interface with DHCP.

This may be required on cloud instances which default to activating only the first network interface.
This is intended to be a very basic role, if you require anything more complicated use the fully-fledge `network` role.

This also reverts commit https://github.com/openmicroscopy/infrastructure/commit/56280081628a416782d9e71cc4532e03ec828722 from https://github.com/openmicroscopy/infrastructure/pull/178 which was a quick-fix for the issue handled in this PR.


joshmoore:
I'm not sure if I've successfully tested this or if I can. The changes make sense, and as I mentioned, I've used the config here for manually configuring NICs which worked fine.

PR #207, opened 2016-12-08
nginx role can optionally setup mainline instead of stable repo (IDR-0.3.1)

manics:
Adds a variable to setup Nginx mainline (currently 1.11) instead of the default stable (1.10). This does not support switching between these two repositories on an existing Nginx server.

The new IDR caching config requires support for complex maps which requires Nginx 1.11.

### DIFF
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -59,6 +59,7 @@ Vagrant.configure(2) do |config|
 
   [
     "nginx",
+    "nginx-mainline",
     "omero-server",
   ].each do |server|
     config.vm.define "#{server}" do |node|
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -58,6 +58,7 @@ Vagrant.configure(2) do |config|
   end
 
   [
+    "nginx",
     "omero-server",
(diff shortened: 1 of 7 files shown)


joshmoore:
This is probably a pretty good candidate for a test since likely we will not use one of the two paths regularly.

manics:
@joshmoore I've added some tests.

joshmoore:
```
TASK [command] *****************************************************************
ok: [nginx]

TASK [assert] ******************************************************************
ok: [nginx]

RUNNING HANDLER [nginx : restart nginx] ****************************************
changed: [nginx]

nginx                      : ok=8    changed=5    unreachable=0    failed=0

```

for both code paths. :+1:

Note: `--provider openstack` doesn't appear to be working after the recent update:

```
The box you're attempting to add doesn't support the provider
you requested. Please find an alternate box or use an alternate
provider. Double-check your requested provider to verify you didn't
simply misspell it.

If you're adding a box from HashiCorp's Atlas, make sure the box is
released.

Name: centos/7
Address: https://atlas.hashicorp.com/centos/7
Requested provider: [:openstack]
```

PR #208, opened 2016-12-08
Add optional nginx_proxy_cache_key_map for modifying cache key (IDR-0.3.1)

manics:
Adds a parameter to optionally modify the cache-key based on a regex-match.

This is required for the new IDR caching config.

Running this on an existing server will result in a configuration change, but there should be no functional difference (it'll create a map with a single default value which is equivalent to using that value directly).


joshmoore:
Merging to also get gh-209 in. Deployment of demo31 seems to be working fine. Adjustments can come in a follow-up if necessary.

PR #209, opened 2016-12-08
New IDR nginx caching config to ignore _=\d+ query param (IDR-0.3.1)

manics:
https://github.com/openmicroscopy/openmicroscopy/pull/4985 disables client side caching of AJAX requests by adding `_=$TIMESTAMP` as a query parameter. This means we must ignore any query parameter of the form `_=\d+` when checking the cache.

--depends-on #207
--depends-on #208

### DIFF
--- a/ansible/idr-playbooks/group_vars/proxy-hosts.yml
+++ b/ansible/idr-playbooks/group_vars/proxy-hosts.yml
@@ -4,6 +4,10 @@
 
 ######################################################################
 # nginx and SSL
+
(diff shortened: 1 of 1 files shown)


joshmoore:
Deployment of demo31 seems to be working and no KEYs appear with `_=`:

```
[root@demo31-proxy nginx]# find . -type f -exec grep -aE ^KEY {} \; | grep _=
[root@demo31-proxy nginx]#
```

Issue #193, opened 2016-11-23
Fix omero-server systemd service dependencies


manics:
A problem noticed when rebooting an OMERO.server with a non-local postgres and NFS mount:
Networking is stopped before the omero service, which causes omero to hang. This dependency needs to be fixed.
https://github.com/openmicroscopy/infrastructure/tree/master/ansible/roles/omero-server

manics:
Fixed in https://github.com/openmicroscopy/infrastructure/pull/212

PR #212, opened 2016-12-09
test omero-server ice 36, fix systemd service files (IDR-0.3.1)

manics:
Test OMERO.server on ice35 and ice36.
ice35 uses the latest release, ice36 uses the latest dev build so the `omero-user` role can be tested (requires https://github.com/openmicroscopy/openmicroscopy/pull/4906). This should be changed to use the latest release once it's released.

Also includes a fix (hopefully) for https://github.com/openmicroscopy/infrastructure/issues/193


joshmoore:
Quicks questions:

 * Are there any concerns from anyone re: dropping support for 5.2.x? This is only in for the testing of the repositories so I'm inclined to say it's fine. (Dropping support for _running_ 5.2 seems inadvisable)
 * How does one test gh-193?

manics:
> How does one test gh-193?

Reboot an omero server which uses an external postgres. If the network goes down before omero then OMERO.server will hang since it'll lose its DB connection (you should be able to tell from the console logs).

joshmoore:
Spun up a full stack on ome101 to test. `-omero` server rebooted quickly with nothing apparent in the OS console:

```
[[32m  OK  [0m] Stopped OMERO.server.
         Stopping LSB: Bring up/down networking...
[[32m  OK  [0m] Stopped LSB: Bring up/down networking.
...
```

PR #199, opened 2016-11-25
Add haproxy to requirements.yml, add example (IDR-0.3.1)

manics:
This depends on https://github.com/openmicroscopy/ansible-role-haproxy.git which doesn't currently exist, so for initial testing you'll have to change that line in `requirements.yml` to https://github.com/manics/ansible-role-haproxy (currently in process of being transferred to @joshmoore)

PR #213, opened 2016-12-12
Add haproxy to requirements.yml, add test (IDR-0.3.1) 

manics:
Rebased from https://github.com/openmicroscopy/infrastructure/pull/199 (re-opening wasn't possible).


joshmoore:
The requirement change worked well for setting up adhoc servers. More regular testing of the vagrant files should be forthcoming.

PR #214, opened 2016-12-12
Add role for managing /etc/hosts in openstack instances (IDR-0.3.1)

manics:
This role takes a list of openstack ansible groups, and populates `/etc/hosts` on a given server with the IPs and hostnames of all servers in those groups. It may also work for normal (non-openstack) groups too.

See #215 for an example

### DIFF
--- a/ansible/roles-dev/hosts-populate/README.md
+++ b/ansible/roles-dev/hosts-populate/README.md
@@ -1,0 +1,21 @@
+Hosts Populate
+==============
+
+Populates `/etc/hosts` with static addresses for a list of hostgroups.
+
+
+Role Variables
+--------------
+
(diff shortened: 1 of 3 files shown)


joshmoore:
Interesting idea. Looking at it, my first thought was, "this should probably start off life under roles-dev" and it has! I've certainly wished for DNS-esque behavior from the openstack instances, but if this suffices for the moment, :+1:

Note: I could also see referencing `$ENV-database` from `$ENV-omero` rather than using the IP address a la `docker --link PG:db`. I'm certainly nervous when using `psql -h $IPADDRESS` since it's easy to not realize what server you're currently working on, especially when production is still mutable.

PR #215, opened 2016-12-12
Setup /etc/hosts on bastion hosts (IDR-0.3.1)

manics:
This populates `/etc/hosts` on the IDR bastion servers with a list of hosts in the same idr-environment. This means you can `ssh` into the bastion server and connect to other hosts by hostname instead of having to lookup the IP.

e.g.
```
ssh -A `floating-ip-of-demoNN-bastion-host
ssh demoNN-omero
```

--depends-on #214


joshmoore:
Merging for 0.3.1-rc2

PR #216, opened 2016-12-12
Install parallel on IDR OMERO servers (IDR-0.3.1)

manics:
I've no idea why `analysis-tools` was installing `moreutils-parallel`. Changed to install `parallel` and this roles has been added to `idr-playbooks/idr-omero.yml`.


joshmoore:
:+1:

PR #219, opened 2016-12-13
Refactor java role (IDR-0.3.1)

manics:
- Use `_` in variable names
- Control JDK installation with a boolean instead of requiring a list of versions, since it probably makes sense for this to be the same as the JRE versions

This is a breaking change, but nothing is currently using these variables.

### DIFF
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -58,6 +58,7 @@ Vagrant.configure(2) do |config|
   end
 
   [
+    "java",
     "nginx",
     "nginx-mainline",
     "omero-server",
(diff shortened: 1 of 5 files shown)


joshmoore:
Fixed conflict a la https://help.github.com/articles/resolving-a-merge-conflict-on-github/

joshmoore:
For @openmicroscopy/devops discussion: concerns regarding this being a breaking change?

joshmoore:
cF. https://galaxy.ansible.com/openmicroscopy/java/ -- happy to migrate this PR to galaxy as java 2.0.0.

joshmoore:
Merging. If anyone needs the old behavior, they can use the galaxy role 1.0.0. I'll migrate this as 2.0.0, and then we can look into consuming it.

NB: would be good to find a way to make the test pass on multiple provision steps in the long run.

PR #223, opened 2016-12-14
Initialise all NICs when multiple cloud networks are present (IDR-0.3.1)

manics:
Attempt to initialise all NICs before any other configuration is done.

This is important when using a bastion server on multiple networks, since by default only the first NIC is initialised, which would prevent internal hosts from being accessed via the uninitialised NIC.


joshmoore:
:+1:

PR #224, opened 2016-12-14
Miscellaneous small IDR fixes (IDR-0.3.1)

manics:
- `management-hosts` shouldn't be referenced directly, it's already part of an idr-environment group
- `Changed a variable test so that the proxy can be automatically configured in the absence of an analysis platform`
- `Changed the EBI NFS mount options to allow auto-negotiation`
- Updated the `os-idr-snapshot.sh` script.


joshmoore:
:+1:

PR #225, opened 2016-12-14
Separate role to reboot server (IDR-0.3.1)

manics:
A development role for rebooting servers either if the kernel has been changed, or unconditionally, depending on the configuration options.

Also adds an IDR playbook for rebooting servers after a kernel upgrade.

Once this is stable it should replace the reboot tasks in `roles/upgrade-distpackages`.

This role/playbook are needed because rebooting servers behind a proxy host gets complicated if the proxy host also requires a reboot, since it'll lead to the unexpected interruption of a SSH connections to internal servers. This means it must be carefully ordered.


joshmoore:
Merging as a dev role.

PR #230, opened 2016-12-20
Openstack re-installation

manics:
Updates to roles whilst re-installing openstack


joshmoore:
Merging as the current state of the UoD OpenStack.

PR #232, opened 2016-12-20
Use unversioned CentOS image

sbesson:
While testing the new OpenStack installation /cc @joshmoore @manics


manics:
Looks good to me

PR #233, opened 2016-12-21
Disable all routes apart from eth0 in network-cloud-interfaces (IDR-0.3.2)

manics:
If a CentOS 7 cloud instance has multiple openstack DHCP interfaces the default gateway/route may end up on a random interface. This causes problems with external access if only one of the interfaces has a floating IP. This PR adds a client-side DHCP hook to disable all routes other than for `eth0`.

This is a dumb "do-the-right-thing" role for setting up cloud networks so it's not configurable. If you need something more complicated use the `network` role in combination with additional tasks for handling config.


joshmoore:
Tested via a fresh spin up of demo32 (named test32). The first time there were some networking issues (i.e. a task failed for dockermanager). Re-running corrected them. A second run after deleting all instances & networks and re-creating ran without error.

PR #234, opened 2016-12-21
Demo32 cleanups (IDR-0.3.2)

manics:
- The post-kernel-upgrade reboot is handled in a separate playbook (not run by default)
- Multi-interface network config is handled in another playbook


joshmoore:
:+1:

PR #236, opened 2016-12-23
Disable nginx port_in_redirect (IDR-0.3.1)

manics:
This is an attempt to make Nginx redirects relative to `host:port` instead of the default which is to use the `host` only. It didn't have the desired effect with OMERO.web, but it's been deployed anyway.

- http://nginx.org/en/docs/http/ngx_http_core_module.html#port_in_redirect
- http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name_in_redirect

### DIFF
--- a/ansible/roles/nginx-proxy/templates/nginx-confd-proxy.j2
+++ b/ansible/roles/nginx-proxy/templates/nginx-confd-proxy.j2
@@ -63,6 +63,8 @@ server {
     proxy_set_header {{ nginx_proxy_forward_scheme_header }} $scheme;
 {% endif %}
 
+    port_in_redirect off;
+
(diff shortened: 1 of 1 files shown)


joshmoore:
_nods_

Can you give an example of a URL that you were looking to have redirect differently?

manics:
`ssh idr-gateway-proxy -L 12345:localhost:80`
```
curl -IL localhost:12345/webclient
HTTP/1.1 302 Moved Temporarily
Server: nginx/1.11.7
Date: Fri, 06 Jan 2017 11:26:15 GMT
Content-Type: text/html
Content-Length: 161
Location: http://localhost/webclient/userdata/?experimenter=-1
Connection: keep-alive

curl: (7) Failed to connect to localhost port 80: Connection refused
```

PR #238, opened 2017-01-05
Add jq install to openstack post

joshmoore:
The snapshotting scripts under ansible/scripts
depend on jq being installed locally.


joshmoore:
Should be fine. Merging.

PR #239, opened 2017-01-06
Use external Ansible roles (round 1+2+IDR+devel) (IDR-0.3.2)

manics:
Replaces `roles/*` and `roles-dev/*` with decoupled galaxy or `ansible-role-*` repository where possible. See `requirements.yml`

Also includes a fix for `.travis` to ensure it fails if `ansible-galaxy` has an error (always seems to exit 0).

## Testing

    git clone https://github.com/openmicroscopy/infrastructure tmp
    cd tmp/ansible/

Fetch `requirements.yml` from this PR

    curl https://raw.githubusercontent.com/manics/infrastructure/external-roles/ansible/requirements.yml > requirements.yml

Delete all roles which are in `requirements.yml`

    for d in $(grep name: requirements.yml | awk '{print $3}'); do rm -rf roles/$d roles-dev/$d; done

Install roles into `vendor/`, then move devel roles into `roles-dev` and the rest to `roles/`, 

    ansible-galaxy install -r requirements.yml
    mv vendor/{hosts-populate,omero-user,reboot-server} roles-dev/
    mv vendor/* roles/

Check for differences

    git status
    git diff

Expected significant differences:
- `java`: https://github.com/openmicroscopy/ansible-role-java/pull/1
- `omero-server`: https://github.com/openmicroscopy/ansible-role-omero-server/commit/72a13151beb626b6b6672c8899da451af1dd51de

In addition this is probably one of the few PRs that should be tested with all vagrant machines.


joshmoore:
Initially, this looks quite sane. Haven't tested locally yet.

manics:
I think this is up to date for all our galaxy roles or repositories.

joshmoore:
Performed the `git status` testing: looks good. Haven't (yet) performed the mega-vagrant test.

joshmoore:
See also: https://trello.com/c/4Ubg1yGK/144-doc-new-role-locations

sbesson:
Briefly discussed, the proposed changes are quite large and effectively correspond to a redesign of our Ansible development process and the way we  work with roles. The expectation is that this migration will satisfy a certain number of needs expressed over the last few months. This is generally in agreement with the the overall process of decoupling, developing & versioning standalone components (roles in this case) happening across the board.

Next obvious step is the documentation of this new workflow (see https://github.com/openmicroscopy/infrastructure/pull/239#issuecomment-273122553). Having received no strong objection from the rest of the sysadmin team, merging this and improvements bug fixes can be carried out in follow-up PRs.

PR #241, opened 2017-01-13
Allow choice of NIC for floating IP when multiple

manics:
This was an attempt to control which network a floating ip is assigned to when an instance has multiple networks. I'm opening this PR as a reminder to consider rebasing to https://github.com/IDR/ansible-role-openstack-idr-instance

--exclude


manics:
See https://github.com/IDR/ansible-role-openstack-idr-instance/issues/1

PR #249, opened 2017-02-14
Clean up IDR inventory and Vagrantfile

manics:
- Removes the out-of-date dynamic inventory `inventory/openstack-private.py`. The correct up-to-date version is in https://github.com/IDR/deployment/pull/6
- Removes the IDR test from the Vagrantfile


jburel:
Will you migrate the other one i.e. ``inventory/openstack.py``

manics:
No, `inventory/openstack.py` should return public floating IPs for openstack instances and isn't used in the IDR. If no-one else is using it we can either remove it, or leave it in case it's useful.

PR #253, opened 2017-03-08
install nodejs 6.x

jburel:
Enable epel-testing to install nodejs 6.x

The change uses epel-testing repo to install nodejs 6.x
This will install v6.9.5

cc @joshmoore @kennethgillen

### DIFF
--- a/ansible/roles/omero-web-apps-build-dependencies/tasks/redhat.yml
+++ b/ansible/roles/omero-web-apps-build-dependencies/tasks/redhat.yml
@@ -20,7 +20,7 @@
 - name: Install nodejs
   become: yes
   yum:
-    enablerepo: epel-testing
+    enablerepo: epel
     name: nodejs
     state: present
 
--- a/ansible/roles/omero-web-apps-build-dependencies/tasks/redhat.yml
+++ b/ansible/roles/omero-web-apps-build-dependencies/tasks/redhat.yml
@@ -20,7 +20,8 @@
 - name: Install nodejs
   become: yes
   yum:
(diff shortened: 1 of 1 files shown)


jburel:
@manics do you know if there is a way to specify to only take nodejs from epel-testing

aleksandra-tarkowska:
Have you tried: ``` yum install <name> --enablerepo=epel-testing```

jburel:
that's what is happening in this role
wanted to make no other things could be install from the repo

manics:
@jburel Ensure the repo is disabled by default in the `.repo` file, and only enable it for a single Ansible task with `enablerepo` http://docs.ansible.com/ansible/yum_module.html

jburel:
Will check the ``.repo`` file

kennethgillen:
@jburel - if you just need "6.x" standard C7 EPEL provides `nodejs-6.9.1`.

I suspect if you try to run `enablerepo: epel-testing` against SLS systems, it won't just work. SLS don't provide a cache of `epel-testing`, so it's not available to the server - it'd have to be an external repo that we add. And if we're going to do that, we may as well just add the upstream node.js.  

If the ansible `enablerepo` translates to ` yum install nodejs --enablerepo=epel-testing`, on a standard SLS configuration:

```
[khgillen@infra-testpr ~]$  sudo yum install nodejs --enablerepo=epel-testing
Error getting repository data for epel-testing, repository not found
```

jburel:
@kennethgillen it translates to the command yes
I will drop the ``epel-testing`` to standard

joshmoore:
Not yet deployed. Intended for cowfish and the spiders. Needs testing.

joshmoore:
Heading to own repo

PR #256, opened 2017-04-04
Update Dell FC630 bios and idrac versions

manics:
Updates  the `dell-update-fc630.yml` playbook with the latest versions of the BIOS and iDRAC. These can be deployed with `ansible-playbook -i <INVENTORY> hardware-dell/dell-update-fc630.yml -l <LIMIT-HOSTS> --ask-become-pass` and will involve an automatic reboot.


joshmoore:
Seems fine.

PR #258, opened 2017-04-10
ansible: Add system update playbook

rleigh-codelibre:
`roles/system-base-update`: Update base operating system; will update FreeBSD, Linux (Debian/Ubuntu, RedHat/CentOS), MacOS X and Windows systems
    
`roles/system-package-update`: Update packages using package manager on systems with a package manager; will update FreeBSD (pkgng), Linux (apt, yum) and MacOS X (homebrew).  Similar to the base update for apt and yum, though apt is more conservative here (safe upgrade only, no new kernel packages etc.).

`system-update`: New playbook which runs both the above roles.

Testing:

You'll need to add to `ansible.cfg`:

```
[defaults]
allow_world_readable_tmpfiles = True
```

to allow unprivileged user switching on MacOS X (deficiency in ACL support in ansible for MacOS X AFAICT).

```
ansible-playbook -l ci-jenkins-osx  -i ~/code/management_tools/ansible/inventory/ci-hosts ~/code/infrastructure/ansible/system-update.yml --vault-password-file=~/code/credentials/ci/vault
ansible-playbook -l ci-jenkins-freebsd  -i ~/code/management_tools/ansible/inventory/ci-hosts ~/code/infrastructure/ansible/system-update.yml --vault-password-file=~/code/credentials/ci/vault
```

The Windows and Linux variants will become supported as we move those nodes over to ansible.


manics:
There's already a Centos 7 upgrade role: https://github.com/openmicroscopy/ansible-role-upgrade-distpackages. Could this be extended?

In practice when would you use `system-base-update` instead of `system-package-update`? Could these roles be combined?

rleigh-codelibre:
The two roles could be combined.

The separation exists in the roles because on some systems e.g. MacOS X with homebrew, there's a big difference between updating the base system and updating homebrew; they are separate actions.  We might want to update homebrew more frequently than the base system (the playbook doesn't do this, but the possibility is there for other playbooks to do so).

snoopycrimecop:
Conflicting PR. Removed from build [MANAGEMENT_TOOLS-merge#1361](https://ci.openmicroscopy.org/job/MANAGEMENT_TOOLS-merge/1361/). See the [console output](https://ci.openmicroscopy.org/job/MANAGEMENT_TOOLS-merge/1361/consoleText) for more details.
Possible conflicts:
  - PR #136 joshmoore 'Add a lookup plugin for `pass`'
    - ansible/.gitignore

snoopycrimecop:
Conflicting PR. Removed from build [MANAGEMENT_TOOLS-merge#1362](https://ci.openmicroscopy.org/job/MANAGEMENT_TOOLS-merge/1362/). See the [console output](https://ci.openmicroscopy.org/job/MANAGEMENT_TOOLS-merge/1362/consoleText) for more details.
Possible conflicts:
  - PR #136 joshmoore 'Add a lookup plugin for `pass`'
    - ansible/.gitignore

rleigh-codelibre:
@manics While we can certainly move this out and merge it with the existing role, it might be helpful to stage it all in this repo first.  This is just one step of what will be many steps to get all this working on 5+ separate platforms.  I don't think splitting it up at the get go will help us do that in an efficient or timely manner.  Could we get to the point where everything is functional before splitting it?

If that's OK, could we merge this so we can start staging all the dependent pieces and get some machines up and running on openstack?  I'd like to be able to get some done this week.

ansible/roles/system-package-update/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# Defaults
```
manics: There's no need to commit empty files

sbesson:
Discussed earlier today with @manics and @kennethgillen. This PR makes sense as a first step and should be considered as development roles/playbooks for now. Merging assuming https://github.com/openmicroscopy/infrastructure/pull/258#discussion_r111115193 can be addressed in the following PR.

PR #259, opened 2017-04-12
ansible: Add generic setup logic, initially for FreeBSD

rleigh-codelibre:
See https://github.com/openmicroscopy/management_tools/pull/371 for testing instructions


ansible/ci-initial-setup.yml
```diff
@@ -0,0 +1,9 @@
+---
+# Initial setup to enable administration with ansible.
+
+- hosts: all
```
rleigh-codelibre: What should be used instead?

rleigh-codelibre: Removing it doesn't work, it tells me it's missing.  How do I restrict to just the specified host?  This is not intended to be run more than once on a single specified host; it's one-time setup to make that host manageable by ansible.

manics: Which ever group is specific to that class of servers... e.g. `ci`, `ci-freebsd`,

ansible/roles/ci-base-setup/tasks/freebsd.yml
```diff
@@ -0,0 +1,12 @@
+---
```
rleigh-codelibre: Use of raw removed.  This was so it could be run without python being installed, but that's (1) insufficient and (2) python is installed but with a specific version, which the ci-jenkins-freebsd group has been set to use by default.

ansible/roles/ci-scratch-storage/tasks/freebsd.yml
```diff
@@ -0,0 +1,60 @@
...
+    line: 'zfs_enable="YES"'
+  become: yes
+  register: rc_zfs
+
+- name: Reload zfs
+  service:
+    name: zfs
+    state: reloaded
+  become: yes
+  when: rc_zfs.changed
```
rleigh-codelibre: Looking at this, the strict ordering requirements make this trick to do with notify.  I've used this elsewhere.

rleigh-codelibre:
@manics All the comments should be addressed now.  The VMs have been recreated with the updated logic.

ansible/roles/ci-base-setup/tasks/freebsd.yml
rleigh-codelibre: This is no longer using `raw`.

ansible/roles/ci-admin-user/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+  user:
+    name: ci-admin
+    comment: "CI Administrator for Ansible use"
+    group: ci-admin
+    password: "{{ ci_admin_password_hash }}"
+    state: present
+  become: yes
+
+- include: freebsd.yml
+  when: ansible_os_family == 'FreeBSD'
```
manics: IIRC last week you wanted to get rid of the "skipped tasks" messages. Does something like `include: "{{ ansible_os_family }}.yml"` work (and create possibly empty `{{ ansible_os_family }}.yml` files for the other supported OSes)?

ansible/roles/ci-data-repo/templates/freebsd-nsmb.j2
```diff
@@ -0,0 +1,10 @@
+# smbfs configuration
+
+[default]
+workgroup=LIFESCI-AD
+
+[FS0]
+addr=fs0.lifesci.dundee.ac.uk
```
manics: Have you thought about making this a variable? It'll be easier if we switch servers in future.

ansible/system-update.yml
```diff
@@ -5,4 +5,3 @@
   roles:
     - role: system-base-update
     - role: system-package-update
-
```
manics: I'll leave it to others to decide on the use and naming of this playbook (`all` implies it's for production systems too).

ansible/roles/pkg-jdk8/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Requirements
+------------
+
+None.
+
+Author Information
+------------------
+
+ome-devel@lists.openmicroscopy.org.uk
```
manics: I'd vote for a longer name, e.g. `java-jdk8` but not a blocker for an internal role, so I'll let others decide.

joshmoore:
Merging. Follow-up cards filed.

PR #263, opened 2017-04-19
Remove deprecated texlive role

sbesson:
With the release of OMERO 5.3.0, nothing in our software stack should depend on texlive for documentation. Removing this role from the infrastructure repository. If some usage shows up, we can always filter branch it and resurrect it as a separate role/repo.

NB: 0263d942aaa83942ddfa4b704692ddd06ac399f8 should already have removed the only usage in our playbooks.

PR #264, opened 2017-04-26
docker: Set sticky flag on /scratch directories

joshmoore:
In order to make /scratch generally usable, it needs to
have similar settings as /tmp.

### DIFF
--- a/ansible/idrsystems-playbooks/idr-analysis.yml
+++ b/ansible/idrsystems-playbooks/idr-analysis.yml
@@ -34,3 +34,9 @@
   - role: openmicroscopy.versioncontrol-utils
 
   - role: openmicroscopy.celery-docker
+
+- hosts: idr-analysis
+  tasks:
+  - file:
+      path: /scratch
+      mode: 1777
--- a/ansible/idrsystems-playbooks/idr-analysis.yml
+++ b/ansible/idrsystems-playbooks/idr-analysis.yml
@@ -34,9 +34,3 @@
   - role: openmicroscopy.versioncontrol-utils
 
   - role: openmicroscopy.celery-docker
-
-- hosts: idr-analysis
-  tasks:
-  - file:
-      path: /scratch
-      mode: "1777"
--- a/ansible/idrsystems-playbooks/idr-analysis.yml
+++ b/ansible/idrsystems-playbooks/idr-analysis.yml
@@ -39,4 +39,4 @@
   tasks:
   - file:
       path: /scratch
-      mode: 1777
+      mode: "1777"
--- a/ansible/idrsystems-playbooks/idr-docker.yml
+++ b/ansible/idrsystems-playbooks/idr-docker.yml
@@ -23,9 +23,3 @@
   - role: openmicroscopy.docker
   - role: openmicroscopy.sudoers
   - role: openmicroscopy.versioncontrol-utils
-
-- hosts: idr-docker
-  tasks:
-  - file:
-      path: /scratch
-      mode: "1777"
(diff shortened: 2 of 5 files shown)


joshmoore:
travis == `successfully: - sorry, openmicroscopy.system-monitor-agent was not found on...`

ansible/idrsystems-playbooks/idr-analysis.yml
```diff
@@ -34,3 +34,9 @@
   - role: openmicroscopy.versioncontrol-utils
 
   - role: openmicroscopy.celery-docker
+
+- hosts: idr-analysis
+  tasks:
+  - file:
+      path: /scratch
+      mode: "1777"
```
manics: We don't need a scratch here on idrsystems

ansible/idrsystems-playbooks/idr-docker.yml
```diff
@@ -23,3 +23,9 @@
   - role: openmicroscopy.docker
   - role: openmicroscopy.sudoers
   - role: openmicroscopy.versioncontrol-utils
+
+- hosts: idr-docker
+  tasks:
+  - file:
+      path: /scratch
+      mode: "1777"
```
manics: We don't need a scratch here on idrsystems

ansible/idrsystems-playbooks/idr-analysis.yml
joshmoore: i.e. the above lvm-partition should be removed or it doesn't need to be 1777?

manics: Doesn't need to be `1777`. Given that we're using scratch on NFS I should probably have removed the partition it from the playbook in the first place.

manics:
As an aside idrsystems-playbooks is almost decoupled from the main infrastructure repo (e.g. it has a separate ansible.cfg and requirements.yml) so we might want to consider our policy on opening cross-directory PRs (or not).

ansible/idrsystems-playbooks/idr-analysis.yml
joshmoore: Shall I remove all reference to `/scratch` then?

manics: Not sure, since it's already there though not being used. I guess either remove it or leave the playbooks unchanged?

joshmoore: Choice is really up to you. Guess if we remove `/scratch` completely, should we use `state: absent` a la https://github.com/debops/ansible-lvm/blob/master/tasks/manage_lvm.yml?

manics: Could do, though up to now I've avoided deleting stuff in roles/playbooks unless necessary, on the grounds that you can never be sure whether that's equivalent to not having added something in the first place (e.g. installing packages automatically brings in deps), so you'd generally want to re-iamge before any big change in the role of a server.

joshmoore:
Reverted.

joshmoore:
Trying to get travis happy.

manics:
Try removing some of the roles from requirements? https://github.com/openmicroscopy/infrastructure/pull/263#issuecomment-295350467

joshmoore:
Well, I had to _add_ this one otherwise the playbook wouldn't run.

manics:
OK...we really need to decouple these directories :)

joshmoore:
I know!

joshmoore:
` [WARNING]: - openmicroscopy.omero-common was NOT installed successfully: -` should now be unrelated. @manics : anything else?

manics:
Thanks

manics:
(I'd merge this but `All required status checks on this pull request must run successfully to enable automatic merging.`)

joshmoore:
We _gotta_ figure out a fix for https://github.com/ansible/galaxy-issues/issues/259

kennethgillen:
Can we have logical branching in the requirements? i.e. we don't want system-monitor-agent if it's not an SLS docker host. At least for now.

joshmoore:
What we've done elsewhere is to have multiple requirements files, e.g. one for with "openmicroscopy." prefixed and one without.

PR #269, opened 2017-05-10
docs: Add image creation instructions

rleigh-codelibre:
Add instructions for creating and uploading OpenStack images for FreeBSD and Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,13 @@ provided in the [ansible README](ansible/README.md).
 
 ## OpenStack
 
-The [OpenStack docs](docs/openstack/idr-openstack.md) outline the installation of OpenStack
-which is running at the University of Dundee.
+- [Server setup](docs/openstack/idr-openstack.md) outlines the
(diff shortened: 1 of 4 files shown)


joshmoore:
Cool, thanks for these, @rleigh-codelibre. Two quick points:
 * I'd think we could skip (or refactor out) the local setup.
 * Can you link them somewhere from higher-up in the hierarchy (I'm trying to keep everything reachable from https://github.com/openmicroscopy/infrastructure/blob/master/README.md)

rleigh-codelibre:
@joshmoore Should be updated now if you want to retry.

joshmoore:
Thanks @rleigh-codelibre. I went ahead and added a link to another file that was dangling (https://github.com/openmicroscopy/infrastructure/pull/269/commits/6bd75ff30897af2919034f775b236ee3535178d8). `Local setup` may eventually need to be combined with `Basic Usage` or `Basic Usage` split, but for the moment, let's call this good. Cheers.

PR #272, opened 2017-05-18
Further ansible roles and playbooks for C++ build slaves

rleigh-codelibre:
- Completes FreeBSD support (can build all existing C++ jobs)
- Completes Debian support (Ubuntu 14.04 and 16.04)
- Implements a large part of the needed Windows support

Running systems can be seen here: http://134.36.162.174:8080/

- freebsd host is barbel
- freebsd-minimal host is bream
- ubuntu-14.04 host is walu
- ubuntu-16.04 host is uaru

(future: centos6 host is cichlid; centos7 host is coelacanth)

Various C++ jobs have been run against both, which can be seen as blue in the job list.

Note the deployment instructions in `docs/openstack/ci-setup.md`.  Duplicate the host configuration for one of the nodes in management_tools (name it after a free IP address in the pool, and also put that IP address in the config), and then follow the rest of the instructions for this node.


joshmoore:
This is obviously a lot to review. So a few initial questions on the roles before I get into the playbooks:

 * Do you have an overview of your roles, their naming scheme, and how you see them evolving?
 * Have you compared the overlap between your roles and existing [org roles](https://galaxy.ansible.com/openmicroscopy/)? Are there any you'd like to deprecate/replace?
 * My understanding from https://trello.com/c/hBMqf7d9/51-extract-roles is that after FreeBSD these roles were going to be extracted. Now Windows is being added. What's the ETA for that extraction? (If you don't see these roles as being intended for extraction & re-use, then let's discuss moving them to a dedicated playbook directory in this PR: http://docs.ansible.com/ansible/playbooks_roles.html)

rleigh-codelibre:
In general:

- roles prefixed with `ci-` are for CI node setup, `ci-jenkins` for jenkins-specific parts of that
- roles prefixed with `pkg-` are for straightforward package installation; these install a single package when Windows has complex installation requirements; some are broader package collections where this restriction doesn't apply
- roles prefixed with `system-` are system maintenance tasks

There are a small number of overlaps with the existing roles, e.g. git installation, jdk installation.  These can likely be merged together in the future once we're happy with what's needed for the CI side, and that the needs of CI and production services are aligned, i.e. we have the flexibility to update things for ongoing CI maintenance without affecting production.  Some parts might benefit from being kept separate for that reason.  For the most part, there's little overlap; the bulk of the roles added here have no counterparts with the existing ones.

Splitting the roles into multiple repositories hasn't been forgotten.  But I would like to be happy that the separation we've achieved here is suitable, and that everyone is generally happy with these roles.  We've not added all the platforms yet, and I'm not fully sure that the organisation is final.  This is still a work in progress, and is coming to be about halfway complete with this PR.  We still need to complete the Windows, CentOS 6, CentOS 7 and MacOS X support to get a complete picture of what we're doing here, and so I would strongly prefer to defer the splitting until we've reached that point.

joshmoore:
I think it's difficult for anyone else to know if they are generally happy with these roles until A) discussion has happened to compare pre-existing roles and what you are trying to achieve and B) the remaining roles are proposed in a reviewable format rather than in very large PRs. I'd also strongly prefer for us to NOT wait until after you've added 4 more platforms because at that point, refactoring will likely be too late.

At this point, I think your options are either to start extracting your roles or to move everything to a subdirectory under the assumption that no playbooks other than the ones in that subdirectory will make use of them.

rleigh-codelibre:
As I mentioned, there are just two roles which overlap that I can see.  The existing roles, being CentOS7-specific, are utterly trivial, and adding CentOS support to pkg-git and pkg-jdk8 will be as simple as copying that logic over as far as I can see with a brief look.

By subdirectory, you mean a subdirectory of `ansible/roles`?  If so, what does that achieve over where they are right now?

joshmoore:
> As I mentioned, there are just two roles which overlap that I can see. The existing roles, being CentOS7-specific, are utterly trivial, and adding CentOS support to pkg-git and pkg-jdk8 will be as simple as copying that logic over as far as I can see with a brief look.

Sorry, but without having discussions about this design, it's not possible to judge what's "utterly trivial". What you are proposing here is new API and what you term as "simply copying logic" is a deprecation and possibly a breaking change. I don't think it's too much to ask that these discussions be had *before* you have implemented this for another 4 platforms.

> By subdirectory, you mean a subdirectory of ansible/roles? If so, what does that achieve over where they are right now?

No. I mean something like `ansible/ci-playbooks/roles` and what that achieves is that then no one can mistake them for general use.

rleigh-codelibre:
Err, I'm not suggesting any breaking changes.  I'm intentionally keeping it separate to avoid interference with the production roles, as requested.  Any unification of the two can happen after we've completed this initial work getting a functional setup.  Doing it as part of this set of PRs seems unnecessarily premature.

I can certainly move them into a separate subdirectory.  That would be both the playbooks *and* the roles?  Would a top-level `ansible-ci` be more appropriate, or just `ansible/ci` if it's both playbook, roles and everything?

joshmoore:
> I'm intentionally keeping it separate to avoid interference with the production roles, as requested.

Sorry if this got expressed somewhere: the reason we have versioned roles is specifically so that nothing can really interfere with them.

> Any unification of the two can happen after we've completed this initial work getting a functional setup. Doing it as part of this set of PRs seems unnecessarily premature.

I think this is just avoiding discussions that should happen earlier rather than later. From my point of view, roles are _the_ unit of re-use in Ansible, I can think of 3-4 different levels:

| Location              | Java-visibility | Implications/Comments              |
|-----------------------|-----------------|------------------------------------|
| your own repo         | private         | "None of our business"             |
| `playbook-dir/roles`  | package-private | "Only for us with these playbooks" |
| `infra/ansible/roles` | ???             | ???                                |
| Roles on galaxy       | public          | "Maintained"                       |

Your creating a number of roles and saying we'll deal with integration later implies technical debt. We'll have more code to review, more to unify, and ultimately more to maintain. For reviewing this, I would appreciate clarity on what's happening with these roles, what version of the production roles they are intended to become, etc.

For me, the `???` above for `ansible/roles` are problematic because the implied use is unclear. They can be consumed by any playbook but are not versioned. This lack of clarity will lead to issues.

> I can certainly move them into a separate subdirectory. That would be both the playbooks and the roles? Would a top-level ansible-ci be more appropriate, or just ansible/ci if it's including playbooks, roles and everything else?

If so, yes, playbooks and roles and no, not `ansible-ci` since there is other infrastructure (especially `ansible.cfg`) which is key for the working of this directory.

ansible/ci-playbooks/roles/ci-base-setup/tasks/redhat.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: ci-base-setup (redhat) | Set hostname
+  hostname:
+    name: "{{ inventory_hostname }}"
```
kennethgillen: @rleigh-codelibre - why are you forcing a hostname here? What if I'm running this on a system with an already-assigned hostname, and use, say, a cname in my inventory? e.g. octopus/eel, trout, etc?

kennethgillen: This possible just needs to be conditionally applied if, say, the hostname matches the hostname from the initial OpenStack image you were cloning from, or something like that.

manics: The hostname is automatically set by cloud-init when you spin up a new instance, there shouldn't be an "initial" name.

ansible/ci-playbooks/roles/ci-jenkins-workspace/tasks/linux.yml
```diff
@@ -0,0 +1,56 @@
+---
+- name: ci-jenkins-workspace (debian) | Add jenkins logical volume
+  lvol:
+    vg: scratch
```
kennethgillen: @rleigh-codelibre - can this be made configurable? At the moment, you're forcing the requirement for another partition on the system via a separate `vg`. I'd personally feel better adding jenkins-scratch to my existing `vg` via a new `lv` only.

ansible/ci-playbooks/roles/ci-jenkins-workspace/tasks/linux.yml
```diff
@@ -0,0 +1,56 @@
+---
+- name: ci-jenkins-workspace (debian) | Add jenkins logical volume
```
kennethgillen: "(debian)"? this is for all `linux`, right?

ansible/ci-playbooks/roles/ci-jenkins-workspace/tasks/linux.yml
```diff
@@ -0,0 +1,56 @@
...
+    lv: scratch
+    size: 64512
+  become: yes
+
+- name: ci-jenkins-workspace (debian) | Create scratch filesystem
+  filesystem:
+    # Because it's a scratch filesytem with no important data on it,
+    # disable journalling for performance.
+    fstype: ext4
+    dev: /dev/scratch/scratch
```
kennethgillen: Definitely needs to be configurable. Can this just use the definition of the `lv` we've already defined above? e.g. could you use ansible-role-lvm-partition, like https://github.com/openmicroscopy/infrastructure/blob/master/ansible/server-state-playbooks/ome-dundeeomero.yml#L130

ansible/ci-playbooks/roles/ci-jenkins-workspace/tasks/linux.yml
rleigh-codelibre: Yes, it's a mistake, fixed now.

kennethgillen:
@rleigh-codelibre - want to add lsb_release in as a dep somewhere? It's pretty handy for Jenkins.

joshmoore:
Summarizing various discussions with @manics, @sbesson, @kennethgillen, @rleigh-codelibre and @jburel :

 * This PR has grown so large that it is unreviewable.
 * There are several outstanding TODOs especially around the deployment of OMERO on C7 (as tested by moving the breaking queue from trout to carp).
 * Some discussions will be needed after the fact (like use of the legacy `/opt/hudson`) which might break hosts which are currently based on these playbooks.
 * Roles contained here may need rationalizing with existing roles on galaxy.
 * That all being said, this PR represents the next major version of CI roles/playbooks ("v2") and to ease collaborative development, this PR needs to be merged in, which I'll do now. I'll also push a `2.0.0-m1` tag and begin the process of migrating non-CI playbooks **out** of this repository making this versioning scheme purely for CI matters.

PR #273, opened 2017-05-19
ome-dundeeomero next steps / cleanup

kennethgillen:
Next steps and cleanup for the New Nightshade server pair.

### DIFF
--- a/ansible/server-state-playbooks/nightshade-web/playbook.yml
+++ b/ansible/server-state-playbooks/nightshade-web/playbook.yml
@@ -45,7 +45,6 @@
     # Upgrade 5.2.8 Web to 5.3.0
     - role: openmicroscopy.omero-web
       omero_web_release: 5.3.2
-      omero_web_upgrade: True
       no_log: true
 
     - role: openmicroscopy.system-monitor-agent
--- a/ansible/server-state-playbooks/nightshade-web/playbook.yml
+++ b/ansible/server-state-playbooks/nightshade-web/playbook.yml
@@ -32,7 +32,7 @@
         name: git
         state: latest
 
-    - name: Omero.web plugin | omero-fpbioimage - plugin install
+    - name: Omero.web plugins | omero-fpbioimage, omero-figure - plugin install
       become: yes
       pip:
         name:
--- a/ansible/server-state-playbooks/nightshade-web/playbook.yml
+++ b/ansible/server-state-playbooks/nightshade-web/playbook.yml
@@ -1,36 +1,7 @@
 # Install OMERO.web with a public user on localhost
 
 - hosts: all
-
-  pre_tasks: 
-
-    # For OMERO.web apps
-    # Pythonpath must exist before omero.web.config is provisioned
-    # or OMERO.web won't start.
-    # create systemd file addition for PYTHONPATH
-    - name: OMERO.web apps | (pre_task) configure systemd for pythonpath - create config folder 
-      become: yes
-      file:
-        dest: "/etc/systemd/system/omero-web.service.d"
-        state: directory
-        mode: "u=rwx,go=rx" 
-        owner: "root" 
-        group: "root" 
-
-    # systemd web-apps folder to pythonpath to env
-    - name: OMERO.web apps | (pre_task) configure systemd for pythonpath 
-      become: yes
-      blockinfile:
-        create: yes
-        destfile: /etc/systemd/system/omero-web.service.d/pythonpath.conf
-        owner: "root" 
-        group: "root" 
-        block: |2+
-          [Service]
-          Environment="PYTHONPATH=$PYTHONPATH:/opt/omero/web/web-extensions"
-      notify:
-        - reload systemd
-
+    
   roles:
 
     # Root LV Size 
@@ -65,9 +36,10 @@
       become: yes
       pip:
         name:
-        #- "omero-figure"
+        - "omero-figure"
         # to be modified, not available on pypi
         - "git+git://github.com/ome/omero-fpbioimage.git@master#egg=fpbioimage"
+        editable: False
         state: present
         # variable comes from role openmicroscopy.omero-web
         virtualenv: "{{ omero_web_basedir }}/venv"
@@ -147,48 +119,9 @@
               }
       notify:
         - restart nginx
-
-    # 'manual' install of omero.web.apps for Nightshade feature parity
-
-    # note: system user var defined in openmicroscopy.omero-web as a default
-    - name: OMERO.web apps | top-level folder
-      become: yes
-      file:
-        path: "{{ omero_web_extensionsdir }}"
-        state: directory
-        owner: "root" 
-        group: "root" 
-        mode: "u=rwx,go=rx" 
-
-    # download figure
-    - name: OMERO.web apps | download latest figure
-      become: yes
-      unarchive:
-        src: https://downloads.openmicroscopy.org/latest/figure.zip
-        owner: "root" 
-        group: "root" 
-        mode: "u=rwx,go=rx" 
-        dest: "{{ omero_web_extensionsdir }}"
-        remote_src: True
-        creates: "{{ omero_web_extensionsdir }}/figure-{{omero_web_extensions_figure_ver}}" 
-
-    # create symlinks
-    - name: OMERO.web apps | app-name symlinks
-      become: yes
-      file:
-        src: '{{ omero_web_extensionsdir }}/{{ item.src }}'
-        dest: '{{ omero_web_extensionsdir }}/{{ item.dest }}'
-        state: link
-        owner: "root" 
-        group: "root" 
-      with_items:
-       - { src: 'figure-{{ omero_web_extensions_figure_ver }}', dest: 'figure' }
-      
-      # put the OMERO.web config for webapps here, and then restart web.
-      # i.e. take it out the initial set of omero.web config
-      # to go into {{ omero_web_basedir }}/config/*.omero
-      # which should then be turned into OMERO.web config by the
-      # omero.web systemd-based restart.
+     
+    # Config for OMERO.web plugins, loaded into OMERO.web by the
+    # omero.web systemd restart.
     - name: 
       become: yes
       template:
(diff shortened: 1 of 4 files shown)


kennethgillen:
Full playbook run of 'nightshade-web' tested on a test system (infra-testpr), installs from pypi omero-figure and FPBioimage viewer.

kennethgillen:
NFS remote processing mount added to playbook and deployed on live system. Tested working.

kennethgillen:
@joshmoore  - note the addition of an `upgrade_webapps == "True"` - cli-flag based upgrade for safety. Else 'state: present'.

kennethgillen:
If any @openmicroscopy/devops would like to have a look-over that'd be good.

I merged to document the changes made to the live system, to continue work on the `ome-demoserver`, but it's still worth an 👀 from others.

ansible/server-state-playbooks/nightshade-web/playbook.yml
```diff
@@ -1,35 +1,16 @@
...
-        owner: "root" 
-        group: "root" 
-        block: |2+
-          [Service]
-          Environment="PYTHONPATH=$PYTHONPATH:/opt/omero/web/web-extensions"
-      notify:
-        - reload systemd
+    # Set desired state of webapps (upgrade v.s. install/keep current version)
+    - set_fact:
+        webapps_state: present
```
manics: FYI `set_fact` will set a host-var that may persist across playbooks in the same `ansible-playbook` command, so there's a (small) danger of inadvertent variable conflicts, especially with shorter names and more complex deployments. Since this var is only used in one place you could get rid of it completely and in `name: Omero.web plugins | plugin install via pip & pypi` have `state: "{{ webapps_state | default('present') }}"`, and pass `-e webapps_state=latest` on the command line to upgrade.

ansible/server-state-playbooks/nightshade-web/playbook.yml
```diff
@@ -42,18 +23,35 @@
       lvm_lvfilesystem: "{{ provision_root_filesystem }}"
 
     # OMERO.web configuration in host_vars in different repository
-    # Upgrade 5.2.8 Web to 5.3.0
     - role: openmicroscopy.omero-web
       omero_web_release: 5.3.2
-      omero_web_upgrade: True
       no_log: true
 
+    # This to be refactored for RHEL/non-spacewalk at some point
```
manics: If you refactor I can think of two options:
- Install/configure `openmicroscopy.system-monitor-agent` in each server's playbook as at present
- Remove all mentions of `openmicroscopy.system-monitor-agent` in individual playbooks and have a single site-wide playbook that handles internal monitoring on all servers.

ansible/server-state-playbooks/ome-dundeeomero/templates/nightly-pg_dump-omero.sh.j2
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+# Clean the files from the folder-format dump directory 
+rm {{ omero_server_db_dumpdir_parent }}/{{ omero_server_db_dumpdir_name }}/* 
+# Create a fresh db dump
+sudo -u postgres pg_dump -j {{ (ansible_processor_count * ansible_processor_cores) |round|int }} -Fd -f {{ omero_server_db_dumpdir_parent }}/{{ omero_server_db_dumpdir_name }} {{ vault_omero_server_dbname }}
```
manics: Why do you need `|round|int`?

ansible/server-state-playbooks/ome-dundeeomero/playbook.yml
```diff
@@ -164,8 +184,28 @@
       command: cp "{{ check_mk_postgres_plugin }}"  /usr/share/check-mk-agent/plugins/
       when: check_mk_postgres_plugin_st.stat.exists
 
+    - name: PostgreSQL Nightly Backups | Create the backups directory
+      become: yes
+      file:
+        path: "{{ omero_server_db_dumpdir_parent }}/{{ omero_server_db_dumpdir_name }}"
```
manics: Minor point, the `omero_server_` prefix is used by the omero-server role, so there's potential for confusion between vars specific to this playbook and those in the role.

ansible/server-state-playbooks/ome-dundeeomero/playbook.yml
```diff
@@ -11,6 +11,27 @@
            ((ansible_virtualization_type is defined)
            and (ansible_virtualization_type == "VMware"))
 
+    # Not required for an OMERO server install - UoD specific implementation
```
manics: Does https://galaxy.ansible.com/openmicroscopy/nfs-mount/ work for this mount?

ansible/server-state-playbooks/nightshade-web/playbook.yml
```diff
@@ -1,35 +1,16 @@
 # Install OMERO.web with a public user on localhost
+# Pass the command line 'upgrade-webapps' flag to upgrade webapps
```
manics: See my later comment, but if you want to keep it as it is there's a typo: `_` not `-`

ansible/server-state-playbooks/nightshade-web/templates/omero-web-config-for-webapps.j2
```diff
@@ -2,7 +2,23 @@
 # Add web-extension OMERO.web configuration
 # after web already installed and running via role
 
-config set -- omero.web.apps '["figure"]'
-config set -- omero.web.ui.top_links '[["Data", "webindex", {"title": "Browse Data via Projects, Tags etc"}], ["History", "history", {"title": "History"}], ["Help", "http://help.openmicroscopy.org/", {"target": "new", "title": "Open OMERO user guide in a new tab"}], ["Figure", "figure_index", {"target": "new", "title": "Open OMERO.Figure in a new tab"}] ]'
```
manics: Minor point: there's no templating in this file (apart from `ansible_managed`) so it could be a normal file and copy.

ansible/server-state-playbooks/nightshade-web/playbook.yml
```diff
@@ -42,18 +23,35 @@
       lvm_lvfilesystem: "{{ provision_root_filesystem }}"
 
     # OMERO.web configuration in host_vars in different repository
```
manics: So `omero_web_config_set` is in hostvars, but `omero_server_config_set` is in `ome-dundeeomero/playbook.yml`?

ansible/server-state-playbooks/ome-dundeeomero/templates/nightly-pg_dump-omero.sh.j2
kennethgillen: Jinja2 was creating doubles - or certainly values with a decimal point, which at least nginx did _not_ appreciate.

ansible/server-state-playbooks/ome-dundeeomero/playbook.yml
kennethgillen: Not sure, didn't see/try it. What is there just now certainly worked.

joshmoore: Fair enough, but that's certainly the point of the reviews is so that we express to the next person coming along here's how _we_ think it should be done.

ansible/server-state-playbooks/nightshade-web/playbook.yml
kennethgillen: I've found a better way which will work with RHEL/non-spacewalk servers - by checking for the SLS nameservers in the networking config - this was introduced on the `ome-dundeeomero` work and continues. These are good ideas for further refactoring though, so thank you.

ansible/server-state-playbooks/nightshade-web/playbook.yml
kennethgillen: As I became more confident with the Ansible, I realised I could define the vars in the playbook, and keep as much as possible public, only pulling the individual private variables from private config. Initially I thought I'd have to keep the whole `omero_web_config_set` together as one. `ome-dundeeomero` represents my having learned that lesson. A refactor of the ns-web playbook would be nice, in this style.

ansible/server-state-playbooks/nightshade-web/playbook.yml
kennethgillen: Yep - thanks, I couldn't find how to do make that conditional just work, without registering the fact. I read that indeed it would persist, so made sure to initialise it first with no conditional to `present`. Thanks for the syntax, and I'll refactor this as you suggest.

ansible/server-state-playbooks/ome-dundeeomero/playbook.yml
kennethgillen: Agreed - I definitely do see the value of these reviews - I didn't skip review because I thought otherwise - rather to progress `ome-demoserver` and have all my ansible code in one (new) branch such that I could diff / merge from my previous code.

ansible/server-state-playbooks/nightshade-web/templates/omero-web-config-for-webapps.j2
joshmoore: My 2c: I don't mind all files being templates.

kennethgillen: It certainly helps when merging >1 playbook with some plain copies and some templating. A 'file' is a template which just doesn't change very much?

ansible/server-state-playbooks/ome-dundeeomero/templates/nightly-pg_dump-omero.sh.j2
manics: Is that because you were divding by 2? https://github.com/openmicroscopy/infrastructure/blob/bb9eff03b7e18452424266a32e0d1691a94d0b23/ansible/server-state-playbooks/nightshade-web/playbook.yml#L71

ansible/server-state-playbooks/nightshade-web/templates/omero-web-config-for-webapps.j2
manics: Yes, unless you have a string that contains `{{` or `{%`

kennethgillen: Good tip, thanks.

PR #275, opened 2017-05-23
Move and update web-proxy playbook

manics:
This moves the web-proxy playbook out of `idrsystems-playbooks` and into `server-state-playbooks/web-proxy/`.


joshmoore:
:+1: for moving the playbook. Some of the other changes concern me since it's essentially working in the opposite direction of `site.yml` -- i.e. less and less able to assert the state of multiple servers at once in the same way. I can see the potential benefit of having a separate vendors directory to prevent one version bump from breaking another, but I'm worried about the maintainability of this.

joshmoore:
Re-reviewing per standup: any thoughts to my previous comments? Perhaps split out the changes if you want the agreed on ones to go in sooner rather than later?

manics:
@joshmoore How's this?

joshmoore:
That's certainly more in line what I was thinking (until we have to do something more drastic). I can try to test via site.yml tomorrow if you're not in a hurry for this to get in.

joshmoore:
I clearly failed to do this. I'm merging and will still test via a separate PR to site.yml

PR #277, opened 2017-06-14
Drop ice-multi-config.sh

joshmoore:
*If* this still works, it will be CentOS6 only. Since
we have moved everythong to CentOS7, this can be removed.


joshmoore:
Note: https://ci.openmicroscopy.org/view/Mgmt/job/ICE-centos6-archive/ can likely be deleted once this is merged (though it may currently point to https://github.com/jburel/ome-sysadmin-scripts/tree/ice-multi-zip/ice-multi-archive) cc: @jburel

joshmoore:
No objections to merging today.

joshmoore:
See also https://github.com/openmicroscopy/zeroc-ice-py-centos7

PR #282, opened 2017-07-10
Ara: add plugin to default configuration

joshmoore:
If ara is not installed, this should produce no change in behavior.

If ara *is* installed, then all calls to `ansible-playbook` will be
recorded to an sqlite database in ~/.ara/.

If additionally, ARA_DATABASE is configured to point at a shared
database, then all runs of ansible-playbook will be centrally recorded.


manics:
What exactly does ara log? Are there any security implications for the data that's stored or output from running ansible?

joshmoore:
The parameters look to be pretty well filtered (and more can be configured to be elided). If anything is printed to stdout, then likely that will be leaked.

PR #284, opened 2017-07-11
www-dev_next-steps

kennethgillen:
This PR to include next steps for www-dev nginx config


sbesson:
Tested against `infra-testpr.openmicroscopy.org` by redeploying the playbook. All worked as expected. Do we need to reprovision from scratch?

kennethgillen:
It might be a good plan at one point, but not necessary yet I'd have thought. I was planning to continue adding config here for redirects.

sbesson:
:+1: Merging then.

PR #286, opened 2017-07-13
Website: second round of redirects

sbesson:
This PR should:

- use the latest released roles instead of custom branches
- consume the merge code for testing analytics /cc @joshmoore 
- uses the 404 page to handle undefined redirects
- add a new set of missing redirects reported during testing /cc @pwalczysko


kennethgillen:
👍  I'd propose merging this all as a basis for more polishing work continuing tomorrow morning.

sbesson:
Fine with getting this in as the current testing state. We can follow-up with adjustements following the secound round of testing. Merging to allow the deployment work on `ome-www` to get started.

PR #287, opened 2017-07-14
Use prod Jekyll configuration files with new jekyll_build roles

sbesson:
This PR removes the need for the manual symlinking of the built content by using an extra `_prod.yml` configuration file overriding the `baseurl`. This should also enable the latest Analytics code in the tested deployment.

Depends on:
- https://github.com/openmicroscopy/www.openmicroscopy.org/pull/92
- https://github.com/openmicroscopy/ansible-role-jekyll-build/pull/5


kennethgillen:
Thanks Seb

PR #289, opened 2017-07-14
Add example of testing for community content

manics:
An example of simple checks of the content of some pages. I'm not planning on adding any more, but @pwalczysko may want to add his own instead of manual testing.


kennethgillen:
👍 - thanks @manics

pwalczysko:
> @pwalczysko may want to add his own instead of manual testing.

@manics : would you have a minute tomorrow to get me into this please ? I am not sure I understand the structure and scope

PR #290, opened 2017-07-14
WWW cleanup

sbesson:
See https://trello.com/c/HZ3WNrmX/19-move-vault-lookup-to-the-inventory. Following the first round of testing, this PR performs a few cleanup actions:

- renames a few remaining places as `www` instead of `www-dev`
- switches back to the http://github.com/openmicroscopy/www.openmicroscopy.org repository for deploying the website
- removes the `vault` lookup step from the public playbook
- removes the deprecated Jinja template superseded by the usage of `openmicroscopy.nginx-proxy`


kennethgillen:
👍 - works, tested this branch against `ome-www-dev`

PR #293, opened 2017-07-18
ome-www deployment updates

kennethgillen:
updates to resize the LVM lvs and filesystems
also configured jekyll-build with a site name

### DIFF
--- a/ansible/server-state-playbooks/www/requirements.yml
+++ b/ansible/server-state-playbooks/www/requirements.yml
@@ -1,8 +1,5 @@
 ---
 
-- name: openmicroscopy.nginx
-  src:  https://github.com/openmicroscopy/ansible-role-nginx.git
-
 - name: openmicroscopy.nginx_proxy
   src: https://github.com/openmicroscopy/ansible-role-nginx-proxy.git
   version: 1.4.0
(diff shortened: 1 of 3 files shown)


kennethgillen:
Supporting config required: https://github.com/openmicroscopy/management_tools/pull/492

ansible/server-state-playbooks/www/requirements.yml
```diff
@@ -16,9 +13,6 @@
   src: https://github.com/openmicroscopy/ansible-role-jekyll-build.git
   version: 1.2.0
 
-- name: openmicroscopy.lvm-partition
```
sbesson: Is `lvm-partition` pulled as a dependency of another role? What happens if a deployment includes a version < 1.1.0?

ansible/server-state-playbooks/www/www.yml
```diff
@@ -58,6 +84,7 @@
       jekyll_build_git_repo: "https://github.com/openmicroscopy/www.openmicroscopy.org"
       jekyll_build_force_rebuild: True
       jekyll_build_config: ['_config.yml', '_prod.yml']
+      jekyll_build_name: www.openmicroscopy.org
```
sbesson: One more commit commit to reflect this change in https://github.com/openmicroscopy/infrastructure/blob/master/ansible/server-state-playbooks/www/www.yml#L212

ansible/server-state-playbooks/www/requirements.yml
kennethgillen: It won't be used by a sub-role I'd have thought, given it makes system-wide changes. Certainly not here.

sbesson:
Deployed successfully with @kennethgillen. Overall :+1: for the changes and using non-default deployment locations. Merging as the state of the live deployment.

ansible/server-state-playbooks/www/requirements.yml
manics: It's used by the docker role if you choose the custom storage option https://github.com/openmicroscopy/ansible-role-docker#custom-storage

PR #294, opened 2017-07-19
Redirects test extension

sbesson:
Following a discussion with @pwalczysko and @kennethgillen, this PR backports some of the redirects testing into the Python file

- moves the `tests` directory up one level to be under the same folder as the playbook
- updates the `HOST` variable to test the production website
- adds more example of tests including redirects to the legacy website

To test this PR:

```
cd ansible/server-state-playbooks/www
pip install -r tests/requirements.txt
py.test tests/check_redirects.py
```

NB: in the mid-run, this might be turned into a post-deployment task of the playbook so that it gets checked after each execution - see http://docs.ansible.com/ansible/test_strategies.html


pwalczysko:
The behaviour is as expected as tested following the test in the header of this PR. 
Ready to merge.

PR #298, opened 2017-08-18
Pin openmicroscopy.postgresql to 2.0.0

joshmoore:
While the PG role undergoes a few milestones for
the new 3.0 version, pin all requirements to the
latest stable, 2.0.0.

See https://github.com/openmicroscopy/ansible-role-postgresql/pull/6


manics:
Looks fine from my point of view.

PR #299, opened 2017-08-22
Clean up deprecated Jekyll playbooks

sbesson:
These playbooks were used for the initial website work and are now superseded
by the content under server-state-playbooks/www.


joshmoore:
uod-web is also in use by the vagrant file. Otherwise, this looks good. (The only mention of "jekyll" in MT is of "jekyll-build" and the only mention of "uod-web" is in a README recording a run.)

sbesson:
:+1: pushed the extra removal

PR #302, opened 2017-09-15
Initial provisioning of pg-int-testing

kennethgillen:
Creation of a new `server-state-playbook` for pg-int-testing to host PostgreSQL databases internally to be used for testing of some kind or another, as opposed to hosting databases for external or internal production systems.

Inventory: https://github.com/openmicroscopy/management_tools/pull/540

No doubt this could be refactored - we can redeploy the system and refactor the playbook later - I need it now, so created with this playbook.

Some fixes to come, e.g. renaming pg_dump vars (since it's not an OMERO server) and probably updating that in inventory etc too.

### DIFF
--- a/ansible/server-state-playbooks/pg-int-testing.yml
+++ b/ansible/server-state-playbooks/pg-int-testing.yml
@@ -1,0 +1,117 @@
+# Install PostgreSQL and prepare the OME (UoD/SLS) prerequisites
+
+- hosts: pg-int-testing.openmicroscopy.org
+  pre_tasks:
+    - name: Install open-vm-tools if system is a VMware vm
+      become: yes
+      yum:
+        name: open-vm-tools
+        state: latest
+      when: >
+           ((ansible_virtualization_type is defined)
+           and (ansible_virtualization_type == "VMware"))
+
+    # Perhaps alter the role at https://github.com/openmicroscopy/ansible-role-lvm-partition/
+    # to make some of the variables non-required.
+    - name: Resize root FS without altering mount options
+      tags: lvm
+      become: yes
+      lvol:
+        lv: root
+        vg: VolGroup00
+        size: "{{ provision_root_lvsize }}"
+        shrink: no
+
+  roles:
+
+    # Disk Layout - PostgreSQL | data dir on separate VG (SSD) 
+    - role: openmicroscopy.lvm-partition
+      tags: lvm
+      lvm_lvname: pgdata
+      lvm_vgname: "{{ provision_postgres_vgname }}"
+      lvm_lvmount: /var/lib/pgsql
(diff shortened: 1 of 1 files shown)


kennethgillen:
Also, this'll need some work to the `pg_hba.conf` no doubt. Possibly that's just different vars given to the role `openmicroscopy.postgresql` we'll find out.

ansible/server-state-playbooks/pg-int-testing.yml
```diff
@@ -0,0 +1,113 @@
...
+      tags: monitoring
+      stat:
+        path: "{{ check_mk_agent_plugin_path }}/mk_logwatch"
+      register: check_mk_logwatch_plugin_st
+
+    - name: Check_MK logwatch plugin | activate the plugin
+      tags: monitoring
+      become: yes
+      command: cp "{{ check_mk_agent_plugin_path }}/mk_logwatch" /usr/share/check-mk-agent/plugins/ creates=/usr/share/check-mk-agent/plugins/mk_logwatch
+      when: check_mk_logwatch_plugin_st.stat.exists
```
manics: Why would these plugins not exist if?

kennethgillen: To stop the whole playbook failing when
* The agent install didn't work for some reason
* If there's a change to the packaging of the agent

Some of the earlier systems don't have all the plugins, so I think point 2 may be relevant in that case.

kennethgillen: I can also see that bit of code being a function or put into the upstream role which accepts `plugin-name`, but that's for a refactor / ☔️  day

manics: This is a new playbook though, so shouldn't everything be working?

kennethgillen: It'd take longer to deploy a system if one were to whittle down working/tested code to only that which "should" be necessary every time. Every change introduces the chance of error, right?

manics: Ansible should give you a well-defined system. If there's uncertainty due to interactions between multiple roles or playbooks this should be highlighted as an issue (and we can decide whether to treat it as a known issue or a bug which needs fixing, as it may cause problems when someone else deploys a new system.

sbesson:
See also https://github.com/openmicroscopy/infrastructure/pull/303. This might have to be reopened against https://github.com/openmicroscopy/prod-playbooks

joshmoore:
Next step here?

kennethgillen:
Shutting this down. Will have to re-open.

kennethgillen:
re-opened on https://github.com/openmicroscopy/prod-playbooks/pull/121

PR #304, opened 2017-09-26
ci-linux-jenkins: Install devtoolset-4

rleigh-codelibre:
For installing devtoolset-4 on cowfish.


joshmoore:
Merging for deployment as part of `ci_v1`. We will need to handle the fact that once we unify the v1 and v2 playbooks in the single repository that there will be a duplication. That might also be the time to start moving to galaxy.

PR #311, opened 2018-03-02
Update gpfs doc with mmnfs for ces share management

manics:
Update gpfs doc


joshmoore:
Will be interesting to see if the addition of a protocol can be done on the fly (i.e. with downtime)

joshmoore:
`/uod/idr-scratch` updated to be read-write with nfs vers 3 on orca-5 & orca-6. cc @jburel

PR #312, opened 2018-04-05
ansible: Install devtoolset-7 on cowfish

rleigh-codelibre:
Description is in the title.  This is exactly the same as the existing devtoolset-4 role, but for devtoolset-7.

### DIFF
--- a/ansible/roles/devtoolset-7/tasks/main.yml
+++ b/ansible/roles/devtoolset-7/tasks/main.yml
@@ -2,7 +2,7 @@
 - name: devtoolset-7 | Install devtoolset-7
   yum:
     name: "{{ item }}"
-    state: latest
+    state: present
   with_items:
     - centos-release-scl
     - devtoolset-7
--- a/ansible/roles/devtoolset-7/tasks/main.yml
+++ b/ansible/roles/devtoolset-7/tasks/main.yml
@@ -7,3 +7,11 @@
     - centos-release-scl
     - devtoolset-7
   become: yes
+
+- name: devtoolset-7 | Remove Java packages which break Jenkins
+  yum:
+    name: "{{ item }}"
+    state: absent
+  with_items:
+    - rh-java-common-runtime
+  become: yes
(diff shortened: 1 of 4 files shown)


ansible/roles/devtoolset-7/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: devtoolset-7 | Install devtoolset-7
+  yum:
+    name: "{{ item }}"
+    state: latest
```
manics: This should be `present` upgrades can be handled separately.

manics:
If the current playbooks are actually broken then perhaps it's best to remove the broken bits?

rleigh-codelibre:
They weren't broken in the past.  Is this due to using a newer ansible version--do the roles need updating?

ansible/roles/devtoolset-7/tasks/main.yml
rleigh-codelibre: I've fixed this now.

rleigh-codelibre:
For the broken jenkinsslave playbook, looks like the files it's trying to copy like `config` and `snoopycrimecop_github` and `snoopycrimecop_github.pub` don't exist in infrastructure, management_tools or credentials.

@sbesson or @joshmoore do you know where these files are and why they aren't in git?

In the meantime, can we just run the individual playbook and avoid this defect?

manics:
I've deployed:

```diff
diff --git a/ansible/ci-deployment.yml b/ansible/ci-deployment.yml
index 806c78e..467c0ba 100644
--- a/ansible/ci-deployment.yml
+++ b/ansible/ci-deployment.yml
@@ -43,6 +43,7 @@
   roles:
     - role: devtoolset-4
     - role: devtoolset-7
+  tags: [temp-tag]

 # Deploy Documentation prerequisites
 - hosts: ci-docs
```
`ansible-playbook -i ../../ansible/inventory/ci-hosts ci-deployment.yml --diff -l cowfish.openmicroscopy.org --ask-become-pass -t temp-tag -v`

We should figure out what's going on with the rest of the playbook.

manics:
I've deployed the latest commit.

PR #316, opened 2020-05-01
Add scc, omego and devspace as infrastructure submodules

sbesson:


joshmoore:
Is your intent that this repository because _solely_ the super parent for infrastructure submodules?

(If so, I could also see including omero-test-infra and build-infra)

sbesson:
That fits with what I had in mind for this repository. Another submodule candidate might be https://github.com/ome/ansible-roles.

